PR #2, opened 2019-04-12
Introduce vhost crate to support vhost-kernel and vhost-user backend drivers

jiangliu:
This PR migrates vhost_backend and vhost_gen code from the firecracker project into the vhost crate.
Support of vhost-user protocol will be added later.


src/lib.rs
```diff
@@ -1,7 +1,168 @@
...
+    ///
+    /// # Arguments
+    /// * `queue_index` - Index of the queue to modify.
+    /// * `fd` - EventFd that will be signaled from guest.
+    fn set_vring_kick(&self, queue_index: usize, fd: &EventFd) -> Result<()>;
+}
+
+/// An interface for controlling vsock vHost backend devices.
+#[cfg(feature = "vhost-vsock")]
+pub trait VhostVsock<'a>: VhostDevice<'a> {
```
sboeuf: We could define this one in its own `vsock.rs` file.

And also, why do you need this to be a `trait`? Do you expect multiple implementation of this?

sys_util/src/eventfd.rs
```diff
@@ -0,0 +1,140 @@
...
+use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
+use std::{io, mem, result};
+
+use libc::{c_void, dup, eventfd, read, write, EFD_NONBLOCK};
+
+/// A safe wrapper around a Linux eventfd (man 2 eventfd).
+///
+/// An eventfd is useful because it is sendable across processes and can be used for signaling in
+/// and out of the KVM API. They can also be polled like any other file descriptor.
+pub struct EventFd {
```
sboeuf: What's the status on this crate `vmm-sys-utils`? I'm asking because we cannot merge the PR as it is now, right? I don't think we want to pull `sys_utils/...` into the `vhost` crate.

src/vhost_kern/vhost_binding.rs
```diff
@@ -0,0 +1,405 @@
...
+//
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the THIRD-PARTY file.
+
+/* Auto-generated by bindgen then manually modified for simplicity */
+
+#![allow(non_upper_case_globals)]
```
sboeuf: This is being discussed on the ML, and this is not clear yet, but it seems you will have to put those bindings in a separate crate.

src/lib.rs
```diff
@@ -1,7 +1,168 @@
...
+/// Result of vHost operations
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// An interface for setting up vhost-based virtio devices.
+///
+/// Vhost-based virtio devices are different from regular virtio devices because the the vhost
+/// backend takes care of handling all the data transfer. The device itself only needs to deal with
+/// setting up the the backend driver and managing the control channel.
+#[cfg(feature = "vhost-device")]
+pub trait VhostDevice<'a>: std::marker::Sized {
```
sboeuf: Thinking about the future with `vhost-user`, do you expect `vhost-user` devices to implement this trait directly (given that it needs to be extended)? Or do you expect the creation of a new `trait` `VhostUserDevice` that would inherit from this trait?

sys_util/src/eventfd.rs
jiangliu: Yes, that's a problem. This crate depends on the vmm-sys-utils and vm-memory.

src/vhost_kern/vhost_binding.rs
jiangliu: I have replied to the ML thread. For the vhost case, I do feel it's better to include the binding into the crate with manual modifications.

src/lib.rs
jiangliu: I think it would be better that vhost-user implements the VhostDevice trait, though we need to extend the VhostDevice trait. Otherwise VhostDevice trait is almost useless:)

src/lib.rs
jiangliu: Sure, I will move it to vsock.rs.
The existing vsock implementation is used a sample, because net and blk is not ready yet.

src/vhost_kern/vhost_binding.rs
sboeuf: @sameo @bonzini @andreeaflorescu @zachreizner need feedback here :)

src/lib.rs
bonzini: Yes, I would expect at least two implementations for vhost-kernel and vhost-user respectively.

sboeuf: `VhostDevice` trait is needed if you want to implement both `vhost_vsock` and `vhost_net`, right?

But in case of vhost-user, I'm think we should create a trait extending `VhostDevice`, as it will be cleaner and clearer from any vhost-user device to implement a trait specific to vhost-user.

bonzini: Maybe rename this to VhostBackend?

sboeuf:
@jiangliu @bonzini @andreeaflorescu what's the status here?
I'm very interested in seeing this crate moving forward in order to base some vhost-user implementations off of it.

jiangliu:
> @jiangliu @bonzini @andreeaflorescu what's the status here?
> I'm very interested in seeing this crate moving forward in order to base some vhost-user implementations off of it.

I'm refine the crate to better support reconnection and almost ready, plan to send out early next week. I have also decoupled Unix domain socket client/server from host-user master/slave roles to enable the       use case that the vhost-user master acts as servers.

Cargo.toml
```diff
@@ -1,8 +1,23 @@
 [package]
-name = "vhost"
+name = "vhost_rs"
```
sboeuf: Why did you change for `vhost_rs`?

jiangliu: it conflicts with firecracker/devices/src/virtio/vhost, so a temp solution to workaround the conflict.

src/vhost_kern/vhost_binding.rs
andreeaflorescu: From the coverage test perspective is better to have the bindings in a different crate. In Firecracker we are excluding auto-generated files when computing the coverage. If we have them in the same crate it means we will have to make changes to the test_coverage.py so that it is more  flexible and we can exclude independent files depending on the repository being tested. The context is the plan to have common integration tests in one repository and all other repositories to consumes them.

But if we don't use most of the things that are auto-generated then maybe we should just have a minimal vhost_bindings file inside the repsository. I don't have any preference between these two.

sboeuf:
@jiangliu the PR LGTM! Are we waiting for review from some folks or do you still have some work planned on this PR?
/cc @andreeaflorescu @bonzini @sameo

jiangliu:
> @jiangliu the PR LGTM! Are we waiting for review from some folks or do you still have some work planned on this PR?
> /cc @andreeaflorescu @bonzini @sameo
Currently waiting reviews, and then need to refine documentation and other required of the rust-vmm project.

src/vhost_kern/vhost_binding.rs
jiangliu: Good point about the test_coverage.py. Could we solve this issue by adding more test cases?

Cargo.toml
bonzini: What is the issue exactly?

src/backend.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub mmap_offset: u64,
+    /// Optional file diescriptor for mmap
+    pub mmap_handle: RawFd,
+}
+
+/// An interface for setting up vhost-based backend drivers.
+///
+/// Vhost-based virtio devices are different from regular virtio devices because the the vhost
+/// backend takes care of handling all the data transfer. The device itself only needs to deal with
+/// setting up the the backend driver and managing the control channel.
```
stefanha: This comment is not generally applicable and relates to vhost-net only.  It may cause confusion.

The comment is essentially saying that in vhost-net the rx/tx virtqueues are handled by vhost and the ctrl virtqueue is handled by the VMM.

In the general case the responsibility of which virtqueue is handled by vhost and which is handled by the VMM is decided when the first vhost implementation of a device is designed.  Usually the dataplane is handled by vhost and anything else is handled by the VMM, but this is not a strict rule or enforced in any way.

I think a clearer way of explaining vhost is that a vhost device is a subset of a virtio devices.  Sometimes all virtqueues but often just those considered the dataplane.  The purpose of vhost is to implement a subset of a virtio device's functionality outside the VMM process.

src/lib.rs
```diff
@@ -1,7 +1,109 @@
...
+//
+// Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE-BSD file.
+
+//! Virtio Vhost Backend Drivers
+//!
+//! Virtio devices use virtqueues to transport data efficiently. Virtqueue is a set of three
+//! different single-producer, single-consumer ring structures designed to store generic
+//! scatter-gather I/O.
```
stefanha: This is outdated as of VIRTIO 1.1.  The new Packed Virtqueue layout is a single ring.

This comment could be updated to:
Virtqueues are often implemented as VIRTIO 1.1 Split Virtqueues with a set of three ...

Eventually this crate should support Packed Virtqueues too for better performance.

stefanha:
> @sboeuf @jiangliu It seems that the item max_queue_num in node should not be configured by VMM. max_queue_num is obtained from slave by calling get_queue_num, to indicate the maximum queue number slave could support. It could be a gatekeeper to check if VMM setup the queues more than slave cloud handle. I see the set_vring_* functions is handling like this. Thus, I thought the support added here should be removed. Correct me if I'm wrong :-P

This appears to be wrong.  Here is what I've found:

In vhost-user "multiple virtqueues" and "multiqueue" are two separate concepts.  Devices with multiple virtqueues have always been possible.  From the beginning of vhost the vhost-net device had an rx virtqueue and a tx virtqueue.  There was just no way for the master to ask the slave if more virtqueues are supported.

Devices with fixed virtqueue layouts are NOT required to advertise VHOST_USER_PROTOCOL_F_MQ because the VMM already knows the number of virtqueues.

When multiqueue devices were implemented it was necessary to extend the protocol so that the slave could report how many virtqueues are available.  Multiqueue devices have a variable number of virtqueues that depends on the capabilities of the device (i.e. the vhost-user slave).

I think that ideally all vhost-user device backends should implement VHOST_USER_PROTOCOL_F_MQ and VHOST_USER_GET_QUEUE_NUM so the master can fetch max_queue_num from the vhost device backend.

However, according to the vhost-user specification and how existing devices are implemented, it is not possible to rely on this today.  Here is how it works:

1. The VMM configures max_queue_num based on the number of queues offered by its virtio device (NOT the vhost device).
2. If the vhost-user device backend supports VHOST_USER_GET_QUEUE_NUM then max_queue_num can be fetched and updates its value.
3. If the VMM user explicitly requested a specific multiqueue number (e.g. on the command-line or through a management API) then device creation fails if this number exceeds max_queue_num (e.g. "Cannot create virtio-net device with 1024 virtqueues, vhost-user slave only supports 64").

Therefore Sebastien's patch or something similar is necessary.  The current master code conflates multiqueue with multiple virtqueues when they are actually separate concepts.

I have sent QEMU patches to make libvhost-user always advertise VHOST_USER_PROTOCOL_F_MQ but please do not rely on this.

bjzhjing:
@stefanha There is one thing I would to confirm with you, that max_queue_num represents the queue pair number, rather than the total number of virtqueues, right?

stefanha:
> @stefanha There is one thing I would to confirm with you, that max_queue_num represents the queue pair number, rather than the total number of virtqueues, right?

max_queue_num is the number of virtqueues.

jiangliu:
> Great, I'll let you know when this will be solved on the virtio-fs side of things.

I plan to integrate this workaround into vhost with comments stating that it's a temporary workaround.

src/lib.rs
jiangliu: The comment has been updated as:
Virtio devices use virtqueues to transport data efficiently. The first generation of virtqueue
is a set of three different single-producer, single-consumer ring structures designed to store
generic scatter-gather I/O. The virtio specification 1.1 introduces an alternative compact
virtqueue layout named "Packed Virtqueue", which is more friendly to memory cache system and
hardware implemented virtio devices. The packed virtqueue uses read-write memory, that means
the memory will be both read and written by both host and guest. The new Packed Virtqueue is 
preferred for performance.

jiangliu:
>In vhost-user "multiple virtqueues" and "multiqueue" are two separate concepts. Devices with multiple >virtqueues have always been possible. From the beginning of vhost the vhost-net device had an rx >virtqueue and a tx virtqueue. There was just no way for the master to ask the slave if more virtqueues are >supported.
Thanks for the explanation, it's easy to mix up these two concepts.
I will carry the patch from @sboeuf in next PR.

src/backend.rs
jiangliu: The comment has been updated as:
/// Vhost devices are subset of virtio devices, which improve virtio device's performance by
/// delegating data plane operations to dedicated IO service processes. Vhost devices use the
/// same virtqueue layout as virtio devices to allow vhost devices to be mapped directly to
/// virtio devices.
/// The purpose of vhost is to implement a subset of a virtio device's functionality outside the
/// VMM process. Typically fast paths for IO operations are delegated to the dedicated IO service
/// processes, and slow path for device configuration are still handled by the VMM process. It may
/// also be used to control access permissions of virtio backend devices.

sameo:
This is superseded by PR #9

PR #11, opened 2020-12-09
Prepare for supporting vhost-net

jiangliu:
This is a misc patch series, which includes:
1) fixes one bug in vring address handling
2) upgrade to latest vm-memory
3) fix a typo
4) enlarge VHOST_USER_MAX_VRINGS to 0x8000

### DIFF
--- a/src/vhost_kern/mod.rs
+++ b/src/vhost_kern/mod.rs
@@ -13,7 +13,7 @@
 
 use std::os::unix::io::{AsRawFd, RawFd};
 
-use vm_memory::{Address, GuestAddress, GuestMemory, GuestUsize};
+use vm_memory::GuestAddressSpace;
 use vmm_sys_util::eventfd::EventFd;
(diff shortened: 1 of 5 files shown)


src/vhost_kern/mod.rs
```diff
@@ -49,15 +48,21 @@ pub trait VhostKernBackend<'a>: AsRawFd {
...
         if queue_size > config_data.queue_max_size
             || queue_size == 0
             || (queue_size & (queue_size - 1)) != 0
         {
-            false
-        } else if GuestAddress(config_data.desc_table_addr)
+            return false;
+        }
+
+        // TODO: the GuestMemory trait lacks of method to look up GPA by HVA,
```
sameo, line 58: What would this lookup method look like? Either the HVA is mapped into the guest physical address space and gpa=hva, or else it'd return an error?

jiangliu: I think it would be something like :
```
fn get_gpa_by_hva(&self, hva: u64) -> Result<GuestAddress>
```

sameo (approved):

PR #24, opened 2021-02-25
Loosen reply ack

stsquad:
This fixes a problem I came across with QEMU and vhost.rs having slightly different ideas about when replies should be handled. We could probably do more to detect mismatches between the reply needed field and the negotiated state of the protocol.

### DIFF
--- a/src/vhost_user/slave_req_handler.rs
+++ b/src/vhost_user/slave_req_handler.rs
@@ -777,7 +777,6 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
         let vflag = VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits();
         let pflag = VhostUserProtocolFeatures::REPLY_ACK;
         if (self.virtio_features & vflag) != 0
-            && (self.acked_virtio_features & vflag) != 0
             && self.protocol_features.contains(pflag)
             && (self.acked_protocol_features & pflag.bits()) != 0
         {


stsquad:
I could do with adding a test for the protocol negotiation but I'm not sure how best to go about it. The connection.rs tests seem to have messages being sent but how does it get passed to the slave?

src/vhost_user/slave_req_handler.rs
```diff
@@ -537,7 +537,6 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
         let vflag = VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits();
         let pflag = VhostUserProtocolFeatures::REPLY_ACK;
         if (self.virtio_features & vflag) != 0
-            && (self.acked_virtio_features & vflag) != 0
```
jiangliu, line 540: It would be great to document the right behavior and why we do not follow the right behavior, to follow a qemu "feature":)

stsquad: Well as far as I'm aware the canonical reference for vhost-user is kept in the QEMU source tree and part of it's [interoperability documentation](https://qemu.readthedocs.io/en/latest/interop/vhost-user.html#protocol-features). I think the key is the wording for the GET/SET protocol features:

    ``VHOST_USER_GET_PROTOCOL_FEATURES``
      :id: 15
      :equivalent ioctl: ``VHOST_GET_FEATURES``
      :master payload: N/A
      :slave payload: ``u64``

      Get the protocol feature bitmask from the underlying vhost
      implementation.  Only legal if feature bit
      ``VHOST_USER_F_PROTOCOL_FEATURES`` is present in
      ``VHOST_USER_GET_FEATURES``.

    .. Note::
       Slave that reported ``VHOST_USER_F_PROTOCOL_FEATURES`` must
       support this message even before ``VHOST_USER_SET_FEATURES`` was
       called.

    ``VHOST_USER_SET_PROTOCOL_FEATURES``
      :id: 16
      :equivalent ioctl: ``VHOST_SET_FEATURES``
      :master payload: ``u64``

      Enable protocol features in the underlying vhost implementation.

      Only legal if feature bit ``VHOST_USER_F_PROTOCOL_FEATURES`` is present in
      ``VHOST_USER_GET_FEATURES``.

    .. Note::
       Slave that reported ``VHOST_USER_F_PROTOCOL_FEATURES`` must support
       this message even before ``VHOST_USER_SET_FEATURES`` was called.

To me the note kind of implies that the negotiation of the protocol is independent of the final SET_FEATURES state. I think if anything we would want to add more clarity to the reference document.

jiangliu: Thanks, seems I missed the "Note" section when implementing the vhost crate:(

jiangliu: could you please help to update the PR? We have just merged a big PR, so rebase is needed:)

stsquad: Will do, just send a clarifcation patch to the vhost-user spec to be super explicit first.

slp (commented):
@stsquad Do we need `vhost_user: add fmt::Debug for VhostUserU64` to be part of this PR? If not, I think we should drop it, rebase and merge `slave_req_handler: relax reply_ack_flag test` ASAP to fix the compatibility with QEMU. Thanks!

stsquad:
> @stsquad Do we need `vhost_user: add fmt::Debug for VhostUserU64` to be part of this PR? If not, I think we should drop it, rebase and merge `slave_req_handler: relax reply_ack_flag test` ASAP to fix the compatibility with QEMU. Thanks!

Done and pushed. I'd also like to get some feedback on better wording on the vhost-user spec because I think @mstsirkin's comments on:

    Subject: [PATCH v2] vhost-user.rst: add clarifying language about protocol negotiation
    Date: Wed,  3 Mar 2021 14:50:11 +0000
    Message-Id: <20210303145011.14547-1-alex.bennee@linaro.org>

imply that even though protocol features must be negotiated before the SET features you may still disable them after the fact. I'd like the final wording to unambiguous either way!

jiangliu (approved):

jiangliu:
@stsquad Could you please help to rebase the MR? We are near to publishing vhost crate now.

slp (approved):

stefano-garzarella (approved):

sboeuf (approved):
I'm fine with this change but honestly this shows how the `vhost-user` spec is loose and not properly defined, leading to different interpretations :(
QEMU is the only proper representation of the spec and it should be specified somewhere that QEMU's code is the reference.

slp:
> I'm fine with this change but honestly this shows how the `vhost-user` spec is loose and not properly defined, leading to different interpretations :(
> QEMU is the only proper representation of the spec and it should be specified somewhere that QEMU's code is the reference.

Yes, that's true. I hope we'll get to fix it, eventually.

stsquad:
There have been some attempts to clarify the language in the docs, for example:

  https://patchew.org/QEMU/20210325144846.17520-1-hi@alyssa.is/

PR #26, opened 2021-03-03
Add custom pipeline yaml for wider clippy coverage

### LINKED ISSUE #25: Check non-default feature code and tests in vhost-ci/clippy-x86
The current `clippy-x86` builder in vhost-ci doesn't cover code for non-default features (e.g. `vhost-kern`, `vhost-user`, etc). So, we have some clippy errors there now, which will be fixed by #23.
To improve code quality, it would be nice if we can pass more flags in clippy there like `clippy --all-targets --all-features`.

keiichiw:
Fixes #25.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,17 @@
+# Copyright 2021 The Chromium OS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE-BSD-Google file.
+
+steps:
+ - label: "clippy-x86-custom"
+   commands:
+    - cargo clippy --all-features --all-targets --workspace -- -D warnings
+   retry:
+    automatic: false
+   agents:
+    platform: x86_64.metal
+    os: linux
+   plugins:
+    - docker#v3.0.1:
+       image: "rustvmm/dev:v11"
+       always-pull: true


keiichiw:
@andreeaflorescu, can I ask you to review this?
It seems that the new check wasn't triggered, but I'm not sure if it's expected. How can we check that my YAML is correct?

andreeaflorescu:
@keiichiw we need to enable the pipeline in Buildkite as well. I would suggest to first review the PR, and then update Buildkite to pick up the changes.

keiichiw:
@andreeaflorescu It makes sense. Then, could you review this PR? Also, who do you know I should ask to update Buildkite?

jiangliu:
Should we also add steps for release buildings for gnu and musl?

keiichiw:
It seems that vm-memory pipelines tests gnu and musl with `backend-mmap`, whose behavior may change depending on libc implementation. However, I don't think the vhost crate uses libc in such a way.
Also, the vhost uses vm-memory with `backend-mmap` feature in dev-dependencies, but release building shouldn't use it. So, I feel we don't have to add such steps. What do you think?

jiangliu (approved):

jiangliu:
@slp @sboeuf Could you please help to review this MR?

sboeuf (commented):
This change LGTM but could you get rid of the 3 merge commits by doing a simple rebase?

keiichiw:
I just rebased the commit.

sboeuf (approved):
Thanks @keiichiw

jiangliu (approved):

jiangliu:
> @keiichiw we need to enable the pipeline in Buildkite as well. I would suggest to first review the PR, and then update Buildkite to pick up the changes.

@andreeaflorescu Any doc/manual to enable Buildkite for vhost?

andreeaflorescu:
I added it: https://buildkite.com/rust-vmm/vhost-ci/builds/85#2247d02c-2192-4ca9-a28c-a180c70acfa3

We don't have documentation for that yet :(

PR #27, opened 2021-03-05
vhost_user: Add support for VHOST_USER_PROTOCOL_F_CONFIGURE_MEM_SLOTS

sboeuf:
This PR introduces the support for recently added message types `VHOST_USER_GET_MAX_MEM_SLOTS`, `VHOST_USER_ADD_MEM_REG` and `VHOST_USER_REM_MEM_REG`, coming from the latest `vhost-user` specification, and gated by the protocol feature `VHOST_USER_PROTOCOL_F_CONFIGURE_MEM_SLOTS`.

They are specifically designed for memory hotplug and memory hot-unplug. They are more efficient than using `VHOST_USER_SET_MEM_TABLE` as the backend does not need to remap all regions if a single region is added or removed.
For the particular case of adding new regions, `VHOST_USER_ADD_MEM_REG` overcome the limitation of 8 regions coming from `VHOST_USER_SET_MEM_TABLE`.

### DIFF
--- a/src/vhost_user/message.rs
+++ b/src/vhost_user/message.rs
@@ -479,6 +479,49 @@ impl VhostUserMsgValidator for VhostUserMemoryRegion {
 /// Payload of the VhostUserMemory message.
 pub type VhostUserMemoryPayload = Vec<VhostUserMemoryRegion>;
 
+/// Single memory region descriptor as payload for ADD_MEM_REG and REM_MEM_REG
+/// requests.
+#[repr(C)]
+#[derive(Default, Clone, Copy)]
+pub struct VhostUserSingleMemoryRegion {
+    /// Padding for correct alignment
+    padding: u64,
+    /// Guest physical address of the memory region.
+    pub guest_phys_addr: u64,
+    /// Size of the memory region.
+    pub memory_size: u64,
+    /// Virtual address in the current process.
+    pub user_addr: u64,
+    /// Offset where region starts in the mapped memory.
+    pub mmap_offset: u64,
+}
+
+impl VhostUserSingleMemoryRegion {
+    /// Create a new instance.
+    pub fn new(guest_phys_addr: u64, memory_size: u64, user_addr: u64, mmap_offset: u64) -> Self {
+        VhostUserSingleMemoryRegion {
+            padding: 0,
+            guest_phys_addr,
+            memory_size,
+            user_addr,
+            mmap_offset,
+        }
+    }
+}
+
+impl VhostUserMsgValidator for VhostUserSingleMemoryRegion {
+    fn is_valid(&self) -> bool {
+        if self.memory_size == 0
+            || self.guest_phys_addr.checked_add(self.memory_size).is_none()
+            || self.user_addr.checked_add(self.memory_size).is_none()
+            || self.mmap_offset.checked_add(self.memory_size).is_none()
+        {
+            return false;
+        }
+        true
+    }
+}
+
 /// Vring state descriptor.
 #[repr(packed)]
 #[derive(Default)]
--- a/src/vhost_user/message.rs
+++ b/src/vhost_user/message.rs
@@ -114,10 +114,30 @@ pub enum MasterReq {
     POSTCOPY_END = 30,
     /// Get a shared buffer from slave.
     GET_INFLIGHT_FD = 31,
-    /// Send the shared inflight buffer back to slave
+    /// Send the shared inflight buffer back to slave.
     SET_INFLIGHT_FD = 32,
+    /// Sets the GPU protocol socket file descriptor.
+    GPU_SET_SOCKET = 33,
+    /// Ask the vhost user backend to disable all rings and reset all internal
+    /// device state to the initial state.
+    RESET_DEVICE = 34,
+    /// Indicate that a buffer was added to the vring instead of signalling it
+    /// using the vring’s kick file descriptor.
+    VRING_KICK = 35,
+    /// Return a u64 payload containing the maximum number of memory slots.
+    GET_MAX_MEM_SLOTS = 36,
+    /// Update the memory tables by adding the region described.
+    ADD_MEM_REG = 37,
+    /// Update the memory tables by removing the region described.
+    REM_MEM_REG = 38,
+    /// Notify the backend with updated device status as defined in the VIRTIO
+    /// specification.
+    SET_STATUS = 39,
+    /// Query the backend for its device status as defined in the VIRTIO
+    /// specification.
+    GET_STATUS = 40,
     /// Upper bound of valid commands.
-    MAX_CMD = 33,
+    MAX_CMD = 41,
 }
 
 impl Into<u32> for MasterReq {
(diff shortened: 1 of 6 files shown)


rbradford (commented):
Unit tests?

sboeuf:
>Unit tests?

Yes I wanted to get feedback from @jiangliu first.

src/vhost_user/message.rs
```diff
@@ -114,10 +114,30 @@ pub enum MasterReq {
...
+    /// Sets the GPU protocol socket file descriptor.
+    GPU_SET_SOCKET = 33,
+    /// Ask the vhost user backend to disable all rings and reset all internal
+    /// device state to the initial state.
+    RESET_DEVICE = 34,
+    /// Indicate that a buffer was added to the vring instead of signalling it
+    /// using the vring’s kick file descriptor.
+    VRING_KICK = 35,
+    /// Return a u64 payload containing the maximum number of memory slots.
+    GET_MAX_MEM_SLOTS = 36,
```
jiangliu, line 128: Should we also implement GET_MAX_MEM_SLOTS in companion with ADD/REMOVE_MEM_REG?

src/vhost_user/message.rs
```diff
@@ -114,10 +114,30 @@ pub enum MasterReq {
...
+    /// Update the memory tables by adding the region described.
+    ADD_MEM_REG = 37,
```
jiangliu, line 130: How about renaming as ADD_MEM_REGION/REMOVE_MEM_REGION?

src/vhost_user/message.rs
```diff
@@ -459,6 +479,49 @@ impl VhostUserMsgValidator for VhostUserMemoryRegion {
 /// Payload of the VhostUserMemory message.
 pub type VhostUserMemoryPayload = Vec<VhostUserMemoryRegion>;
 
+/// Single memory region descriptor as payload for ADD_MEM_REG and REM_MEM_REG
+/// requests.
+#[repr(packed)]
+#[derive(Default, Clone, Copy)]
+pub struct VhostUserSingleMemoryRegion {
+    /// Padding for correct alignment
+    pub padding: u64,
```
jiangliu, line 488: Should it be private? No one should access this field.
BTW, the padding field really seems odd:(

src/vhost_user/slave_req_handler.rs
```diff
@@ -62,6 +62,8 @@ pub trait VhostUserSlaveReqHandler {
     fn get_config(&self, offset: u32, size: u32, flags: VhostUserConfigFlags) -> Result<Vec<u8>>;
     fn set_config(&self, offset: u32, buf: &[u8], flags: VhostUserConfigFlags) -> Result<()>;
     fn set_slave_req_fd(&self, _vu_req: SlaveFsCacheReq) {}
+    fn add_mem_reg(&self, region: &VhostUserSingleMemoryRegion, fd: RawFd) -> Result<()>;
+    fn rem_mem_reg(&self, region: &VhostUserSingleMemoryRegion) -> Result<()>;
```
jiangliu, line 66: How about renaming as "remove_mem_region"?

src/vhost_user/message.rs
sboeuf: Yes I think I should implement it as it's part of the `CONFIGURE_MEM_SLOTS` protocol feature.

src/vhost_user/message.rs
sboeuf: Well I think this is better to stick to the vhost-user spec for this kind of naming. It makes the readability with the spec better, hence it helps maintaining the code.

src/vhost_user/message.rs
sboeuf: Oh yes it should be private.
And yes I agree the padding looks so weird :-/

src/vhost_user/slave_req_handler.rs
sboeuf: Yes I can rename these functions.

src/vhost_user/message.rs
slp: I agree, using the same naming as in vhost-user spec helps when you're "grepping" around.

rbradford: Yeh, the other thing I like to do if the name is not really "compatible" with the Rust way is I add a comment with the kernel/C specification name.

src/vhost_user/message.rs
slp: Out of curiosity, I took a look at the relevant code in QEMU, and I still don't understand why we need that padding. Relevant commits in QEMU:

```
f1aeb14b0809e313c74244d838645ed25e85ea63 Transmit vhost-user memory regions individually
3009edff8192991293fe9e2b50b0d90db83c4a89 vhost-user: fix VHOST_USER_ADD/REM_MEM_REG truncation
```

slp:
This PR looks good to me. Just needs some tests to avoid dropping the coverage (as @rbradford already pointed out) and a rebase.

src/vhost_user/message.rs
sboeuf: Well I relied on the spec here: https://qemu.readthedocs.io/en/latest/interop/vhost-user.html#single-memory-region-description

slp: @sboeuf Yes, your implementation is correct, I was just wondering why we needed that in the first place, when it was added to QEMU.

sboeuf:
@jiangliu @rbradford @slp the PR has been updated, PTAL :)

src/vhost_user/message.rs
```diff
@@ -459,6 +479,49 @@ impl VhostUserMsgValidator for VhostUserMemoryRegion {
 /// Payload of the VhostUserMemory message.
 pub type VhostUserMemoryPayload = Vec<VhostUserMemoryRegion>;
 
+/// Single memory region descriptor as payload for ADD_MEM_REG and REM_MEM_REG
+/// requests.
+#[repr(packed)]
+#[derive(Default, Clone, Copy)]
```
rbradford, line 485: `#[repr(packed)]` is wrong. I checked the C code there it's not defined as packed there so this should be `#[repr(C)]` (this might have no impact on the actual layout but it is more correct.)

sboeuf: Sounds good!

rbradford (approved):

src/vhost_user/slave_req_handler.rs
```diff
@@ -417,6 +435,55 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
...
+                self.send_ack_message(&hdr, res)?;
+            }
+            MasterReq::REM_MEM_REG => {
+                if self.acked_protocol_features
+                    & VhostUserProtocolFeatures::CONFIGURE_MEM_SLOTS.bits()
+                    == 0
+                {
+                    return Err(Error::InvalidOperation);
+                }
+                if rfds.is_some() {
```
jiangliu, line 478: This check may be redundant because check_attached_fds() should have already handled this error case.

sboeuf: Good point, let me remove it.

sboeuf:
@jiangliu comment's addressed!

slp (approved):

jiangliu (approved):

Issue #12, opened 2020-12-09
Release version 0.1.0 to crates.io


jiangliu:
There are several projects using this crate, it would be great to publish this crate to crates.io.

andreeaflorescu:
@jiangliu the coverage on the crate is very low (~40%). We typically published the other crates with a coverage of 80-90%. Are you planning on adding more tests?

jiangliu:
I'm going over the code again, and will help to add more test cases:)

jiangliu:
@sboeuf @slp 
We are preparing for publishing to crates.io. 
The following PR should be merged before publishing:
https://github.com/rust-vmm/vhost/pull/24
https://github.com/rust-vmm/vhost/pull/26

Is there anymore tasks we need to handle before publishing?

sboeuf:
@jiangliu
>Is there anymore tasks we need to handle before publishing?

Not that I can think of. Let's publish once we got #24 and #26 merged :)

slp:
> Not that I can think of. Let's publish once we got #24 and #26 merged :)

I agree. We need #24, at least.

slp:
@jiangliu How are going the plans for releasing it to crates.io? Which PRs must be merged before doing that?

stefano-garzarella:
I'm closing this issue since we already released `0.1.0` on creates.io.
Feel free to reopen if it is wrong to close this.

PR #31, opened 2021-03-17
vhost_user: config offset matches qemu

### LINKED ISSUE #29: Offset in VHOST_USER_GET_CONFIG/SET_CONFIG mismatch with qemu
The vhost crate considers offsets between the config offset and length as valid:
https://github.com/rust-vmm/vhost/blob/ee3e8722706c984b3dfe12d3a130e92101b78e8f/src/vhost_user/message.rs#L662
I believe the literal 0x100 is the config offset.

qemu, on the other hand, considers offsets valid from 0 to length, and assumes the offset uses CONFIG_OFFSET as a base.
See get_config in vhost-user.c, it sets config to offset 0 and length to max to cover the entire area.
https://source.codeaurora.org/quic/qemu/qemu/tree/hw/virtio/vhost-user.c?h=qemu.org/master#n2068

This mismatch causes the qemu GET_CONFIG message to be considered invalid by the vhost crate.

I didn't find anything in the spec indicating which is correct:
https://qemu.readthedocs.io/en/latest/interop/vhost-user.html

Deferring to qemu seems like the best thing for compatibility with qemu, but it might break existing users.

dgreid:
Modify the interpretation of the set and get config offset field to
match qemu. The wording in the vhost user spec is ambiguous, bet lets
defer to qemu's implementation so vhost based devices can be used with
qemu as a VMM. Fixes #29.

Signed-off-by: Dylan Reid <dgreid@chromium.org>

### DIFF
--- a/src/vhost_user/message.rs
+++ b/src/vhost_user/message.rs
@@ -657,14 +657,13 @@ impl VhostUserConfig {
 impl VhostUserMsgValidator for VhostUserConfig {
     #[allow(clippy::if_same_then_else)]
     fn is_valid(&self) -> bool {
+        let end_addr = match self.size.checked_add(self.offset) {
+            Some(addr) => addr,
+            None => return false,
+        };
         if (self.flags & !VhostUserConfigFlags::all().bits()) != 0 {
             return false;
-        } else if self.offset < 0x100 {
-            return false;
-        } else if self.size == 0
-            || self.size > VHOST_USER_CONFIG_SIZE
-            || self.size + self.offset > VHOST_USER_CONFIG_SIZE
-        {
+        } else if self.size == 0 || end_addr > VHOST_USER_CONFIG_SIZE {
             return false;
         }
         true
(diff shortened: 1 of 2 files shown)


src/vhost_user/message.rs
```diff
@@ -659,8 +659,6 @@ impl VhostUserMsgValidator for VhostUserConfig {
     fn is_valid(&self) -> bool {
         if (self.flags & !VhostUserConfigFlags::all().bits()) != 0 {
             return false;
-        } else if self.offset < 0x100 {
-            return false;
         } else if self.size == 0
             || self.size > VHOST_USER_CONFIG_SIZE
             || self.size + self.offset > VHOST_USER_CONFIG_SIZE
```
jiangliu, line 664: Seems we should also check for "self.offset >= VHOST_USER_CONFIG_SIZE", otherwise "self.size + self.offset" may wrap around.

dgreid: Good catch, you're correct. Switched to checked_add and added a test.

slp (approved):
LGTM. I think this should be merged before going with #12

jiangliu (approved):

PR #34, opened 2021-04-27
vhost_user: Fix check for SET_VRING_ENABLE

keiichiw:
We should allow to receive SET_VRING_ENABLE request with non-zero index even if
MQ protocol feature is disabled because some device can have multiple queues
without the MQ feature.
e.g. virtio-net means that it supports multiple pairs of tx/rx queues. So, the
slave must support at least one pair of queues regardless of whether MQ is
supported.

### DIFF
--- a/src/vhost_user/slave_req_handler.rs
+++ b/src/vhost_user/slave_req_handler.rs
@@ -400,8 +400,8 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
             }
             MasterReq::SET_VRING_ENABLE => {
                 let msg = self.extract_request_body::<VhostUserVringState>(&hdr, size, &buf)?;
-                if self.acked_protocol_features & VhostUserProtocolFeatures::MQ.bits() == 0
-                    && msg.index > 0
+                if self.acked_virtio_features & VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits()
+                    == 0
                 {
                     return Err(Error::InvalidOperation);
                 }


stefano-garzarella (commented):
I agree on this change, just a comment:
looking at the [spec](https://qemu-project.gitlab.io/qemu/interop/vhost-user.html#master-message-types) there is this statement:
```
VHOST_USER_SET_VRING_ENABLE
...
This request should be sent only when VHOST_USER_F_PROTOCOL_FEATURES has been negotiated.
```
So, should we check if `PROTOCOL_FEATURES` is set?
```rust
if (self.acked_virtio_features & VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits()) == 0
{
    return Err(Error::InvalidOperation);
}
```

keiichiw:
@stefano-garzarella Good idea! I just updated the commit and its description.

stefano-garzarella (approved):

keiichiw:
I just updated the commit description that CI complained.

src/vhost_user/slave_req_handler.rs
```diff
@@ -399,12 +399,12 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
...
             MasterReq::SET_VRING_ENABLE => {
-                let msg = self.extract_request_body::<VhostUserVringState>(&hdr, size, &buf)?;
-                if self.acked_protocol_features & VhostUserProtocolFeatures::MQ.bits() == 0
-                    && msg.index > 0
+                if self.acked_virtio_features & VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits()
+                    == 0
                 {
                     return Err(Error::InvalidOperation);
                 }
+                let msg = self.extract_request_body::<VhostUserVringState>(&hdr, size, &buf)?;
```
jiangliu, line 407: Should we receive the VhostUserVringState message before checking the PROTOCOL_FEATURES flag? Otherwise there will be a malformed message body left in the communication pipe.

keiichiw: Good catch! Thanks.
Fixed.

jiangliu (approved):

stefano-garzarella (approved):

sboeuf (approved):

PR #38, opened 2021-05-07
vhost_user: implement From instead of Into

danielverkamp:
Implementing From<...> provides the equivalent Into<..> automatically,
so this form is preferred.

Fixes the clippy from_over_into warning:

https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into

Signed-off-by: Daniel Verkamp <dverkamp@chromium.org>


jiangliu (approved):

slp:
Thanks for the PR!

PR #39, opened 2021-05-10
vhost_user: Avoid passing around RawFd

### LINKED ISSUE #37: vhost_user: Use File instead of RawFd in VhostUserSlaveReqHandler trait
Currently, some methods in the `VhostUserSlaveReqHandler` trait, such as `set_vring_kick`, takes `RawFd`s as their arguments. However, this design doesn't allow a trait implementor to know whether the given FDs are valid and who its owner is. This means it cannot call `from_raw_fd()` for the given FD safely.

So, it'd be nice if the `VhostUserSlaveReqHandler` trait is updated to pass around `File`s instead of `RawFd`s. Then, the ownership of the underlying FD will become clearer.

keiichiw:
Use `File` instead of `RawFd` in methods of `SlaveReqHandler` to handle
FDs' ownership easily.
Fixes #37.

Signed-off-by: Keiichi Watanabe <keiichiw@chromium.org>

### DIFF
--- a/src/vhost_user/master_req_handler.rs
+++ b/src/vhost_user/master_req_handler.rs
@@ -1,6 +1,7 @@
 // Copyright (C) 2019-2021 Alibaba Cloud. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::fs::File;
 use std::mem;
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::os::unix::net::UnixStream;
@@ -33,9 +34,7 @@ pub trait VhostUserMasterReqHandler {
     }
 
     /// Handle virtio-fs map file requests.
-    fn fs_slave_map(&self, _fs: &VhostUserFSSlaveMsg, fd: RawFd) -> HandlerResult<u64> {
-        // Safe because we have just received the rawfd from kernel.
-        unsafe { libc::close(fd) };
+    fn fs_slave_map(&self, _fs: &VhostUserFSSlaveMsg, _fd: &dyn AsRawFd) -> HandlerResult<u64> {
         Err(std::io::Error::from_raw_os_error(libc::ENOSYS))
     }
 
@@ -50,14 +49,12 @@ pub trait VhostUserMasterReqHandler {
     }
 
     /// Handle virtio-fs file IO requests.
-    fn fs_slave_io(&self, _fs: &VhostUserFSSlaveMsg, fd: RawFd) -> HandlerResult<u64> {
-        // Safe because we have just received the rawfd from kernel.
-        unsafe { libc::close(fd) };
+    fn fs_slave_io(&self, _fs: &VhostUserFSSlaveMsg, _fd: &dyn AsRawFd) -> HandlerResult<u64> {
         Err(std::io::Error::from_raw_os_error(libc::ENOSYS))
     }
 
     // fn handle_iotlb_msg(&mut self, iotlb: VhostUserIotlb);
-    // fn handle_vring_host_notifier(&mut self, area: VhostUserVringArea, fd: RawFd);
+    // fn handle_vring_host_notifier(&mut self, area: VhostUserVringArea, fd: &dyn AsRawFd);
 }
 
 /// A helper trait mirroring [VhostUserMasterReqHandler] but without interior mutability.
@@ -70,9 +67,7 @@ pub trait VhostUserMasterReqHandlerMut {
     }
 
     /// Handle virtio-fs map file requests.
-    fn fs_slave_map(&mut self, _fs: &VhostUserFSSlaveMsg, fd: RawFd) -> HandlerResult<u64> {
-        // Safe because we have just received the rawfd from kernel.
-        unsafe { libc::close(fd) };
+    fn fs_slave_map(&mut self, _fs: &VhostUserFSSlaveMsg, _fd: &dyn AsRawFd) -> HandlerResult<u64> {
         Err(std::io::Error::from_raw_os_error(libc::ENOSYS))
     }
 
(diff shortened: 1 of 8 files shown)


keiichiw:
@jiangliu Could you take a look?

src/vhost_user/slave_req_handler.rs
```diff
@@ -690,24 +682,23 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
     fn check_attached_rfds(
         &self,
         hdr: &VhostUserMsgHeader<MasterReq>,
-        rfds: Option<Vec<RawFd>>,
-    ) -> Result<Option<Vec<RawFd>>> {
+        rfds: &Option<Vec<RawFd>>,
+    ) -> Result<()> {
```
jiangliu, line 686: The change may cause file descriptor leakage , such as call site at line 291. Functionality of "close_rfds()" should be reserved.

src/vhost_user/slave_req_handler.rs
```diff
@@ -610,27 +606,24 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
...
-                let sock = unsafe { UnixStream::from_raw_fd(fds[0]) };
-                let vu_req = SlaveFsCacheReq::from_stream(sock);
-                self.backend.set_slave_req_fd(vu_req);
-                self.send_ack_message(&hdr, Ok(()))
-            } else {
-                Err(Error::InvalidMessage)
-            }
-        } else {
-            Err(Error::InvalidMessage)
+        let mut files = files.ok_or(Error::InvalidMessage)?;
```
jiangliu, line 611: How about adding a helper function to extract the only one fd from Option<Vec<File>>?

src/vhost_user/slave_req_handler.rs
jynnantonix: Manually closing fds is a pretty strong code smell.  We shouldn't be relying on code reviews to catch leaks like this.

It seems to me like the `recv_header()` method should be returning `File`s rather than `RawFd`s.

src/vhost_user/slave_req_handler.rs
keiichiw: Good idea. I added a helper function [`take_single_file()`](https://github.com/rust-vmm/vhost/pull/39/files#diff-7ac2f087f60290ddeaa6195f8dcb4fc3cace310c7c84b31cbbc97174832fedf7R768).

src/vhost_user/slave_req_handler.rs
keiichiw: I replaced `RawFd`s in connection.rs with `File`s as suggested. As a result, I needed to update both the master handler and the slave handler.
While I used `File` at places where we had transferred ownership of `RawFd`, I used `&dyn RawFd` instead at places where ownership wasn't transferred.

src/vhost_user/master_req_handler.rs
```diff
@@ -437,7 +419,7 @@ mod tests {
         });
 
         fs_cache
-            .fs_slave_map(&VhostUserFSSlaveMsg::default(), fd)
+            .fs_slave_map(&VhostUserFSSlaveMsg::default(), &fd)
```
keiichiw, line 422: Although this test passed on my local environment, this line caused a build error on the CI with the following error:
>  the trait `std::os::unix::io::AsRawFd` is not implemented for `i32`

It seems that rust-vmm-ci is using a compiler older than 1.48.0, where [`AsRawFd` was implemented for `RawFd`](https://doc.rust-lang.org/std/os/unix/io/trait.AsRawFd.html#impl-AsRawFd-13).

@andreeaflorescu Is it possible to upgrade the version of cargo for CI?

jynnantonix: If we can't upgrade the toolchain we can wrap this with a newtype:

```rust
struct BorrowedFd<'a>(&'a RawFd);

impl<'a> AsRawFd for BorrowedFd<'a> {
    fn as_raw_fd(&self) -> RawFd {
        *self.0
    }
}
```

src/vhost_user/slave_req_handler.rs
```diff
@@ -645,23 +625,13 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
...
-                }
-            }
-            None => {
-                if !nofd {
-                    return Err(Error::InvalidMessage);
-                }
-            }
+        let file = take_single_file(files);
+
+        if (nofd && file.is_some()) || (!nofd && file.is_none()) {
```
jynnantonix, line 630: minor nit: I find double negatives really hard to read:  "doesn't not have an fd" vs "has an fd".

src/vhost_user/slave_req_handler.rs
```diff
@@ -800,6 +764,15 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
     }
 }
 
+/// Utility function to take the first element from option of a vector of files.
```
jynnantonix, line 767: nit: Should also mention here that this will return `None` if there is more than one file in the Vec.

jynnantonix (approved):

src/vhost_user/slave_req_handler.rs
keiichiw: Added a comment

src/vhost_user/master_req_handler.rs
keiichiw: Okay, I defined this wrapper type in the test function

src/vhost_user/slave_req_handler.rs
keiichiw: declared `has_fd` instead.

lauralt:
Hi, we added a new container version with Rust 1.52.1 (v12) and for this update to be propagated in vhost crate, we should fix the conflicts in [this PR](https://github.com/rust-vmm/vhost/pull/46) and then merge it. Can any of the crate codeowners do that?

keiichiw:
@jiangliu I just rebased my patches on recent changes. Please take a look again!
Also, can you suggest another reviewer who can give me approval?

jiangliu (approved):

slp (approved):

PR #42, opened 2021-06-04
vhost_user: Add header flags support

### LINKED ISSUE #40: Allow `Master` to set flags in the header
We need a way to allow `Master` to send a request with the header containing some specific flags such as `need_reply`.

sboeuf:
Introduce a new method set_flags() in order to let the caller define the
expected set of flags that should be applied to the header for the
following messages.

Fixes #40

### DIFF
--- a/src/vhost_user/master.rs
+++ b/src/vhost_user/master.rs
@@ -84,6 +84,7 @@ impl Master {
                 protocol_features_ready: false,
                 max_queue_num,
                 error: None,
+                hdr_flags: VhostUserHeaderFlag::empty(),
             })),
         }
     }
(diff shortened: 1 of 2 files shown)


src/vhost_user/master.rs
```diff
@@ -125,6 +126,12 @@ impl Master {
 
         Ok(Self::new(endpoint, max_queue_num))
     }
+
+    /// Set the header flags that should be applied to all following messages.
+    pub fn set_flags(&self, flags: VhostUserHeaderFlag) {
```
sboeuf, line 131: @jiangliu maybe I should call it `set_hdr_flags()`, let me know what you think.

jiangliu: yes, set_hdr_flags() is more clear.

sboeuf: Done!

jiangliu (approved):

jiangliu (approved):

eryugey (approved):

sameo (approved):

PR #45, opened 2021-06-08
vhost_user: Fix NEED_REPLY for all messages

sboeuf:
Except the list of commands from which we can expect a reply, all the
other commands should be able to reply with an acknowledgment if being
asked for.

For the commands such as SET_PROTOCOL_FEATURES, which usually happens
before the protocol features negotiation is complete, we should also be
able to reply with an ACK if needed. Indeed, the protocol features
negotiation could have happened earlier, which means we might expect a
reply ACK from GET_FEATURES command if this happens later on.

### DIFF
--- a/src/vhost_user/master.rs
+++ b/src/vhost_user/master.rs
@@ -158,11 +158,9 @@ impl VhostBackend for Master {
     fn set_features(&self, features: u64) -> Result<()> {
         let mut node = self.node();
         let val = VhostUserU64::new(features);
-        let _ = node.send_request_with_body(MasterReq::SET_FEATURES, &val, None)?;
-        // Don't wait for ACK here because the protocol feature negotiation process hasn't been
-        // completed yet.
+        let hdr = node.send_request_with_body(MasterReq::SET_FEATURES, &val, None)?;
         node.acked_virtio_features = features & node.virtio_features;
-        Ok(())
+        node.wait_for_ack(&hdr).map_err(|e| e.into())
     }
 
     /// Set the current Master as an owner of the session.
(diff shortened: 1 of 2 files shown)


sboeuf:
@jiangliu @rbradford PTAL

src/vhost_user/master.rs
```diff
@@ -148,30 +148,24 @@ impl VhostBackend for Master {
...
     fn set_owner(&self) -> Result<()> {
         // We unwrap() the return value to assert that we are not expecting threads to ever fail
         // while holding the lock.
         let mut node = self.node();
-        let _ = node.send_request_header(MasterReq::SET_OWNER, None)?;
-        // Don't wait for ACK here because the protocol feature negotiation process hasn't been
-        // completed yet.
-        Ok(())
+        let hdr = node.send_request_header(MasterReq::SET_OWNER, None)?;
+        node.wait_for_ack(&hdr).map_err(|e| e.into())
```
jiangliu, line 162: Is it safe to wait for ack here？Previously there‘s a comment saying we shouldn't wait for ack here.

sboeuf: Well I know this comment was there but from my understanding of the spec, it's partly wrong.
Looking at both https://github.com/qemu/qemu/blob/master/docs/interop/vhost-user.rst#communication and https://github.com/qemu/qemu/blob/master/docs/interop/vhost-user.rst#vhost_user_protocol_f_reply_ack it seems that only the following list should NOT implement the ACK mechanism:
- `VHOST_USER_GET_FEATURES`
- `VHOST_USER_GET_PROTOCOL_FEATURES`
- `VHOST_USER_GET_VRING_BASE`
- `VHOST_USER_SET_LOG_BASE`
- `VHOST_USER_GET_INFLIGHT_FD`

And for every other command, if `REPLY_ACK` is negotiated along with the presence of `NEED_REPLY`, then it should reply with an ACK.

sboeuf: @jiangliu WDYT?

sboeuf:
@jiangliu could you please review it one more time?

src/vhost_user/master.rs
jiangliu: Some existing vhost-user drivers work in following sequence:
1) master.set_owner()
2) master.get_features()
3) master.set_features()
4) master.set_protocol_features()
5) master.set_vring_call()

The NEED_REPLY protocol feature should be negotiated in step 4, could please help to review Cloud Hypervisor vhost drivers?

jiangliu: My bad, we already have a check `if self.acked_protocol_features & VhostUserProtocolFeatures::REPLY_ACK.bits() == 0` in wait_for_ack(), so it should be OK. It would be great to also check "protocol_features_ready" flag in `wait_for_ack()`.

jiangliu (approved):

src/vhost_user/master.rs
sboeuf: Yes `wait_for_ack()` is a no-op in case `REPLY_ACK` hasn't been negotiated.
I can add a check for `protocol_features_ready` as well, but if you look at the code, any place where we check `acked_protocol_features` we should also check for `protocol_features_ready`, don't you think?
Maybe the best way to fix that is to make `acked_protocol_features` an `Option<u64>`?

sboeuf: Let me know your thoughts on this, I'll fix it with a separate PR.

jiangliu: Sure, it's a separate issue. Let's move on with this MR.

rbradford (approved):

PR #49, opened 2021-06-22
Use File for slave request fd

jynnantonix:
There is nothing in the vhost-user documentation or spec that says this
must be a socket for fs device requests.  It's not even guaranteed that
it is an fd for a UnixStream.

Change the type to a File and leave it up to the backend to convert it
into a more specialized type if necessary.  Ideally we'd use something
like UniqueFd, which provides close-on-drop functionality but doesn't
make any other assumptions, but since we don't have that File is the
next best thing.

Signed-off-by: Chirantan Ekbote <chirantan@chromium.org>

### DIFF
--- a/src/vhost_user/slave_req_handler.rs
+++ b/src/vhost_user/slave_req_handler.rs
@@ -4,13 +4,11 @@
 use std::fs::File;
 use std::mem;
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
-use std::os::unix::net::UnixStream;
 use std::slice;
 use std::sync::{Arc, Mutex};
 
 use super::connection::Endpoint;
 use super::message::*;
-use super::slave_fs_cache::SlaveFsCacheReq;
 use super::{Error, Result};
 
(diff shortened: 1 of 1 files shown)


slp:
Makes sense, but this will break every consumer of this crate, so we need to be careful about when to merge it.

@jiangliu do you think this should be in before or after 0.1.0 gets published?

jiangliu:
> Makes sense, but this will break every consumer of this crate, so we need to be careful about when to merge it.
> 
> @jiangliu do you think this should be in before or after 0.1.0 gets published?

I have a little preference for releasing v0.1 before merging this MR, so other projects depending on vhost could move on.

slp:
> I have a little preference for releasing v0.1 before merging this MR, so other projects depending on vhost could move on.

Okay, I agree.

jiangliu:
@jynnantonix could please help to rebase the MR?

jiangliu:
After a second thoughts, I'm a little curious what's the usage case for this change?

PR #54, opened 2021-07-08
Inflight I/O: Make structs and members public

harshanavkis:
The queue region and descriptor state structs introduced in the previous commits need to be public to be used by other crates.

### DIFF
--- a/src/vhost_user/message.rs
+++ b/src/vhost_user/message.rs
@@ -783,8 +783,8 @@ impl VhostUserMsgValidator for VhostUserFSSlaveMsg {
 
 /// Inflight I/O descriptor state for split virtqueues
 #[repr(packed)]
-#[derive(Clone, Copy)]
-struct DescStateSplit {
+#[derive(Clone, Copy, Default)]
+pub struct DescStateSplit {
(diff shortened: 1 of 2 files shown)


src/vhost_user/message.rs
```diff
@@ -795,8 +795,15 @@ struct DescStateSplit {
     counter: u64,
 }
 
+impl Default for DescStateSplit {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
jiangliu, line 802: it would be better to move current implementation into default(), and let new() call default().

harshanavkis:
@jiangliu I have addressed your comments in the new set of commits.

harshanavkis:
@jiangliu I have addressed your comments in the new set of commits.

slp (approved):

stefano-garzarella (dismissed):

slp:
@jiangliu @eryugey @sboeuf PTAL

eryugey (dismissed):

sboeuf (changes requested):

sboeuf (approved):

slp (approved):

PR #59, opened 2021-07-25
Inflight I/O: Make struct members public

harshanavkis:
Signed-off-by: Harshavardhan Unnibhavi <harshanavkis@gmail.com>

### DIFF
--- a/src/vhost_user/message.rs
+++ b/src/vhost_user/message.rs
@@ -786,13 +786,13 @@ impl VhostUserMsgValidator for VhostUserFSSlaveMsg {
 #[derive(Clone, Copy, Default)]
 pub struct DescStateSplit {
     /// Indicate whether this descriptor (only head) is inflight or not.
-    inflight: u8,
+    pub inflight: u8,
     /// Padding
     padding: [u8; 5],
     /// List of last batch of used descriptors, only when batching is used for submitting
-    next: u16,
+    pub next: u16,
     /// Preserve order of fetching available descriptors, only for head descriptor
-    counter: u64,
+    pub counter: u64,
 }
 
 impl DescStateSplit {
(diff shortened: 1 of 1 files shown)


harshanavkis:
@sboeuf for the inflight I/O implementation, for now, desc_num field needs to be updated. Hence I made all fields public. Would this be alright or do you prefer to write setter methods for each field individually.

src/vhost_user/message.rs
```diff
@@ -786,13 +786,13 @@ impl VhostUserMsgValidator for VhostUserFSSlaveMsg {
 #[derive(Clone, Copy, Default)]
 pub struct DescStateSplit {
     /// Indicate whether this descriptor (only head) is inflight or not.
-    inflight: u8,
+    pub inflight: u8,
     /// Padding
-    padding: [u8; 5],
+    pub padding: [u8; 5],
```
jiangliu, line 791: Do we need to export padding too?

sboeuf (commented):
PR sounds good to me, but I agree with @jiangliu that you shouldn't make the `padding` fields public.

src/vhost_user/message.rs
```diff
@@ -839,25 +839,25 @@ impl QueueRegionSplit {
 #[derive(Clone, Copy, Default)]
 pub struct DescStatePacked {
     /// Indicate whether this descriptor (only head) is inflight or not.
-    inflight: u8,
+    pub inflight: u8,
     /// Padding
-    padding: u8,
+    pub padding: u8,
```
sboeuf, line 844: Ditto about `padding`.

src/vhost_user/message.rs
```diff
@@ -872,27 +872,27 @@ impl DescStatePacked {
...
+    pub old_used_idx: u16,
     /// device ring wrap counter
-    used_wrap_counter: u8,
+    pub used_wrap_counter: u8,
     /// old device ring wrap counter
-    old_used_wrap_counter: u8,
+    pub old_used_wrap_counter: u8,
     /// Padding
-    padding: [u8; 7],
+    pub padding: [u8; 7],
```
sboeuf, line 893: Ditto about `padding`.

harshanavkis:
@jiangliu @sboeuf makes sense to me. I updated the PR.

sboeuf (approved):

src/vhost_user/message.rs
```diff
@@ -872,27 +872,27 @@ impl DescStatePacked {
 #[repr(packed)]
```
jiangliu, line 872: Why `repr(packed)` instead of `repr(c)` here and above?

jiangliu (approved):

PR #63, opened 2021-08-03
Fix set_log_base() implementation for vhost-user

sboeuf:
The SET_LOG_BASE implementation was incomplete as it didn't include the
ability to send the shared memory region information along with the file
descriptor.

This is required to perform proper dirty page logging with a vhost-user
device.

### DIFF
--- a/src/vhost_user/message.rs
+++ b/src/vhost_user/message.rs
@@ -731,6 +731,35 @@ impl VhostUserMsgValidator for VhostUserInflight {
     }
 }
 
+/// Single memory region descriptor as payload for SET_LOG_BASE request.
+#[repr(C)]
+#[derive(Default, Clone)]
+pub struct VhostUserLog {
+    /// Size of the area to log dirty pages.
+    pub mmap_size: u64,
+    /// Offset of this area from the start of the supplied file descriptor.
+    pub mmap_offset: u64,
+}
+
+impl VhostUserLog {
+    /// Create a new instance.
+    pub fn new(mmap_size: u64, mmap_offset: u64) -> Self {
+        VhostUserLog {
+            mmap_size,
+            mmap_offset,
+        }
+    }
+}
+
+impl VhostUserMsgValidator for VhostUserLog {
+    fn is_valid(&self) -> bool {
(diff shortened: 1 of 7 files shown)


rbradford (approved):
There is the `

src/vhost_user/mod.rs
```diff
@@ -416,7 +416,16 @@ mod tests {
         master.set_vring_enable(0, true).unwrap();
 
         // unimplemented yet
-        master.set_log_base(0, Some(eventfd.as_raw_fd())).unwrap();
+        master
```
jiangliu, line 419: should we remove the comments above?

src/vhost_user/message.rs
```diff
@@ -731,6 +731,28 @@ impl VhostUserMsgValidator for VhostUserInflight {
...
+    /// Create a new instance.
+    pub fn new(mmap_size: u64, mmap_offset: u64) -> Self {
+        VhostUserLog {
+            mmap_size,
+            mmap_offset,
+        }
+    }
+}
+
+impl VhostUserMsgValidator for VhostUserLog {}
```
jiangliu, line 754: Should we validate that "mmap_size + mmap_offset" doesn't wrap around?

src/vhost_user/mod.rs
sboeuf: Yes good point.

src/vhost_user/message.rs
sboeuf: I can check  `mmap_size > 0`, but I'm not sure about the wrapping. `mmap_offset` is the offset in the file descriptor, so there's no invalid value there.

jiangliu: If mmap_offset is offset into the `fd` and `mmap_offset + mmap_size` wraps around, the vhost-user server should fail to mmap the area.

sboeuf: Oh you meant wrap around the u64 value, gotcha, I'll add this check :)

sboeuf: Done!

jiangliu (approved):

rbradford (approved):

PR #68, opened 2021-08-09
vhost_user: enforce ByteValued for recv-d structs

alyssais:
Converting arbitrary bytes into an arbitrary Rust value is unsafe.  For example, it's unsafe to create a `String` that isn't valid UTF-8.  But the various internal `recv*` functions didn't restrict their return types enough to enforce this invariant, making them unsafe without being properly marked.

To fix this, we tighten up the bounds of the functions to enforce that their return types are `ByteValued`, meaning that they can only be used to create types that are safe to initialize with arbitrary data such as might be received over a socket.

It's worth asking how these functions could have been unsafe in the first place, since they don't contain any unsafe blocks themselves. The answer is that the functions that `recv` into `iovec`s are also unsafe but not correctly marked.  I'm preparing further patches to fix that up, but it's a lot of work so I've separated out this change in the hope of getting it in first and making the diff for the next one smaller.

This internal tightening shouldn't result in any public API changes.

Signed-off-by: Alyssa Ross <hi@alyssa.is>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2018"
 [features]
 default = []
 vhost-vsock = []
-vhost-kern = ["vm-memory"]
+vhost-kern = []
 vhost-user = []
 vhost-user-master = ["vhost-user"]
 vhost-user-slave = ["vhost-user"]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -14,7 +14,7 @@ edition = "2018"
 default = []
 vhost-vsock = []
 vhost-kern = ["vm-memory"]
-vhost-user = []
+vhost-user = ["vm-memory"]
```
jiangliu, line 17: How about changing "vm-memory" as a mandatory dependency?

jiangliu (approved):

sboeuf (approved):
LGTM

PR #69, opened 2021-08-12
vhost_user: handle recv_with_fds being unsafe

alyssais:
`recv_with_fds` from vmm-sys-util was never safe, but was incorrectly not marked as safe.  Handle it safely in every function where we have enough information to know that calling it will be safe, and propogate the unsafe to any function where we don't.

Fortunately this is quite straightforward now all of the high-level methods have the guarantee that they're returning `ByteValued` types.

Signed-off-by: Alyssa Ross <hi@alyssa.is>

***

This change is necessitated by https://github.com/rust-vmm/vmm-sys-util/pull/135.  If it's merged before the vmm-sys-util dependency is updated to a version with that change, it will produce compiler warnings, so I've marked it as a draft until that PR is merged.

### DIFF
--- a/src/vhost_user/connection.rs
+++ b/src/vhost_user/connection.rs
@@ -313,7 +313,8 @@ impl<R: Req> Endpoint<R> {
             iov_base: rbuf.as_mut_ptr() as *mut c_void,
             iov_len: len,
         }];
-        let (bytes, _) = self.sock.recv_with_fds(&mut iovs, &mut [])?;
+        // Safe because we own rbuf and it's safe to fill a byte array with arbitrary data.
+        let (bytes, _) = unsafe { self.sock.recv_with_fds(&mut iovs, &mut [])? };
         Ok((bytes, rbuf))
     }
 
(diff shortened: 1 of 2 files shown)


jiangliu:
Hi alyssais, could you please help to fix the CI failures?

alyssais:
@jiangliu the CI failures are just coming from the `unsafe` blocks I added not being required yet because we haven't bumped vhost's vmm-sys-util dependency.

Should I change Cargo.toml to have it use the latest vmm-sys-util from git, or would you like to do a vmm-sys-util release including the latest changes that we could update to?

jiangliu:
> @jiangliu the CI failures are just coming from the `unsafe` blocks I added not being required yet because we haven't bumped vhost's vmm-sys-util dependency.
> 
> Should I change Cargo.toml to have it use the latest vmm-sys-util from git, or would you like to do a vmm-sys-util release including the latest changes that we could update to?

Hi @alyssais , we have just published vmm-sys-util v0.9.0, could you please give another try?

src/vhost_user/connection.rs
```diff
@@ -324,7 +325,10 @@ impl<R: Req> Endpoint<R> {
     /// * - SocketRetry: temporary error caused by signals or short of resources.
     /// * - SocketBroken: the underline socket is broken.
     /// * - SocketError: other socket related errors.
```
stefano-garzarella, line 327: What about adding the following comments like in `recv_with_fds()` documentation?

```rust
    /// # Safety
    ///
    /// It is the callers responsibility to ensure it is safe for arbitrary data to be
    /// written to the iovec pointers.
```

alyssais: Oh yes, good idea!

jiangliu (approved):

slp (approved):
LGTM, thanks!

PR #74, opened 2021-08-31
vhost-user: Implement FromRawFd for Listener

slp:
Implement FromRawFd for Listener, using the underlying UnixListener
FromRawFd implementation.

This is useful when the listener socket is inherited as a file
descriptor from a parent (such as libvirt or a jailer).

Signed-off-by: Sergio Lopez <slp@redhat.com>


sboeuf (dismissed):
Looks good!

jiangliu (approved):

stefano-garzarella (approved):

jiangliu (approved):

PR #75, opened 2021-09-01
Dummy PR for testing the new CI

catalindumitru:
With the new script that autogenerates pipelines, we can customize the configuration of the docker plugin for any particular test. Therefore we can add `/dev/vhost-vdpa-0` to the configuration of just the tests that need it (i.e. to `unittests`). To do this, we need to update the Buildkite step:
```bash
DOCKER_PLUGIN_CONFIG='{
    "tests": ["unittests-gnu"],
    "cfg": {
        "devices": [ "/dev/vhost-vdpa-0" ]
    }
}' ./rust-vmm-ci/.buildkite/autogenerate_pipeline.py | buildkite-agent pipeline upload
```
This also allows us to remove `test-vdpa` from the custom pipeline.

PR #33, opened 2021-03-24
add vhost-vdpa support (in-kernel)

stefano-garzarella:
This PR add the vhost-vdpa support (in-kernel).
Since it requires IOTLB message support for DMA map/unmap features, the first patches added that support to vhost-kern (VHOST_IOTLB, VHOST_BACKEND, vhost_msg_vs), let me know if you prefer a separated PR for that.

While writing the tests, I found an issue in the kernel. The patches are upstream [1][1] [2][2] but not landed in stable kernels yet, so I left the vdpa.set_config_call() commented in the test to avoid kernel panics.

I tested with Linux v5.10 and I used vdpa_sim module (net device):
```bash
modprobe vhost-vdpa
modprobe vdpa-sim   # vdpa-sim-net on Linux v5.11+

cargo test --features=vhost-kern,vhost-vdpa
```
Then `/dev/vhost-vdpa-0` is available and you can run the tests.

Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f6bbf0010ba004f5e90c7aefdebc0ee4bd3283b9
[2]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=0bde59c1723a29e294765c96dbe5c7fb639c2f96

PR #82, opened 2021-09-03
Draft: vDPA without skip test commit

stefano-garzarella:
Draft PR without commit that skip vdpa tests if the device is not available to check if everything is okay


stefano-garzarella:
@catalindumitru I think aarch64 image is not building vdpa kernel modules. What we can do?

The #33 contains a commit to skip the vdpa tests in that case, so it's safe and eventually we can leave it that way.

catalindumitru:
Hi @stefano-garzarella, we have previously checked that the device is actually found by adding an assert at the end of one of the tests (in this [commit](https://github.com/rust-vmm/vhost/pull/75/commits/475371f2da95ebafd15e55e54995d6d614e31dd0)). The tests in the Buildkite [build](https://buildkite.com/rust-vmm/vhost-ci/builds/219#4a36e77e-eb64-4168-8a99-44f571434e54) (including the aarch64 ones) seemed to fail because of this assert. We are looking into the issue.

andreeaflorescu:
@stefano-garzarella skipping the tests in case the device is not found might lead to silent failures in the CI. We might want to just have the tests under a feature in case the device is not found locally during development, but in the CI is safer to have it tested.

stefano-garzarella:
@andreeaflorescu yep, make sense.

So, if you and @catalindumitru agree, I'll close this PR, and remove the [commit](https://github.com/rust-vmm/vhost/pull/33/commits/69e76296438196b881ab3dab7928ff5e3644f22a) that skip the test from #33.

The tests are already under `vhost-vdpa` feature, is that okay?

andreeaflorescu:
We first need to identify the reason for the failure :(

Let's wait a bit until we figure out what is going on.

lauralt:
@stefano-garzarella We weren't able to fix the issue for the aarch64 instance, and we need more time. But anyway, it's a problem with our instances, not with the test, so I'm fine as well with skipping the tests for now. Sorry for the late response :(.

lauralt:
Also, looks like if we add `vhost-vdpa` and `vdpa-sim-net` to a 5.11.0 kernel, there won't be any vhost-vdpa-0 device. Any idea why? Is there a problem with this particular kernel version?

stefano-garzarella:
> @stefano-garzarella We weren't able to fix the issue for the aarch64 instance, and we need more time. But anyway, it's a problem with our instances, not with the test, so I'm fine as well with skipping the tests for now. Sorry for the late response :(.

Okay, thanks for the update! I just rebased #33, but CI is failing with a strange behavior (Permission denied removing stuff). How can I restart it?

stefano-garzarella:
> Also, looks like if we add `vhost-vdpa` and `vdpa-sim-net` to a 5.11.0 kernel, there won't be any vhost-vdpa-0 device. Any idea why? Is there a problem with this particular kernel version?

This is strange, for 5.11 should be enough, but I just remembered that starting from 5.12 a new management API is added to handle vDPA sim devices: https://github.com/torvalds/linux/commit/a3c06ae158dd6fa8336157c31d9234689d068d02

So you need to use `vdpa` tool from [iproute2](https://git.kernel.org/pub/scm/network/iproute2)/iproute2.git to add a new device instance:
```bash
sudo modprobe vhost-vdpa
sudo modprobe vdpa-sim-net
./vdpa dev add mgmtdev vdpasim_net name dev1
```

Then you can check in this way:
```
$ ./vdpa dev list     
dev1: type network mgmtdev vdpasim_net vendor_id 0 max_vqs 2 max_vq_size 256
```

lauralt:
> > @stefano-garzarella We weren't able to fix the issue for the aarch64 instance, and we need more time. But anyway, it's a problem with our instances, not with the test, so I'm fine as well with skipping the tests for now. Sorry for the late response :(.
> 
> Okay, thanks for the update! I just rebased #33, but CI is failing with a strange behavior (Permission denied removing stuff). How can I restart it?

Fixed it.

PR #84, opened 2021-09-09
Inflight I/O: Introduce new error enums

harshanavkis:
These error enums are required to handle various errors during
memfd_create, ftruncate and memfd sealing operations.

Signed-off-by: Harshavardhan Unnibhavi <harshanavkis@gmail.com>

### DIFF
--- a/src/vhost_user/mod.rs
+++ b/src/vhost_user/mod.rs
@@ -83,6 +83,12 @@ pub enum Error {
     FeatureMismatch,
     /// Error from request handler
     ReqHandlerError(IOError),
+    /// memfd file creation error
+    MemFdCreateError,
+    /// File truncate error
+    FileTrucateError,
+    /// memfd file seal errors
+    MemFdSealError,
 }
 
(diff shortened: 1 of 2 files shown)


jiangliu (approved):

stefano-garzarella (approved):

src/vhost_user/mod.rs
```diff
@@ -83,6 +83,12 @@ pub enum Error {
     FeatureMismatch,
     /// Error from request handler
     ReqHandlerError(IOError),
+    /// memfd file creation error
+    MemFdCreateError,
+    /// File truncate error
+    FileTrucateError,
+    /// memfd file seal errors
+    MemFdSealError,
```
andreeaflorescu, line 91: Why do these errors need to be defined here if they're going to be used in another crate?

stefano-garzarella: IIUC because they are returned by functions that returns `vhost::vhost_user::Result`. @harshanavkis any thoughts?

PR #87, opened 2021-09-23
Cargo: replace wildcard with 0.5 for serial_test

slp:
crates.io does not allow publishing crates which have wildcards in
their dependencies. We had one for "serial_test", so replace the
wildcard pointing it to the current major/minor ("0.5").

Signed-off-by: Sergio Lopez <slp@redhat.com>


jiangliu (approved):

stefano-garzarella (approved):
The patch LGTM, should we also avoid `>=` for 0.* dependencies?
E.g. vmm-sys-util marked some API unsafe between two 0.* releases and some crates' tag, like v0.1.0 for this one IIRC, doesn't build anymore.

PR #88, opened 2021-09-23
Inflight I/O: Fix the memory layout of queue region header

zhangjaycee:
As defined in the vhost-user protocol[1], the last field of queue region
header should be a zero-sized array. However, current implementation uses
u64, which will cause a wrong inflight I/O memfd memory layout. So, let's
just remove the desc fileds to fix it.

[1]
https://qemu-project.gitlab.io/qemu/interop/vhost-user.html#inflight-i-o-tracking

Signed-off-by: Jiachen Zhang <zhangjiachen.jaycee@bytedance.com>

### DIFF
--- a/src/vhost_user/message.rs
+++ b/src/vhost_user/message.rs
@@ -877,8 +877,6 @@ pub struct QueueRegionSplit {
     pub last_batch_head: u16,
     /// Idx value of used ring
     pub used_idx: u16,
-    /// Pointer to an array of DescStateSplit entries
-    pub desc: u64,
 }
 
 impl QueueRegionSplit {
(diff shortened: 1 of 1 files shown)


stefano-garzarella:
Yep, it seems wrong. Maybe we should use `vmm_sys_util::fam` for that field.

jiangliu:
could you please help to rebase or try FamStruct from vmm_sys_util crate?

sboeuf (dismissed):

stefano-garzarella:
@zhangjaycee we would like to fix this problem. Do you have plans to complete this PR?

luigix25:
I was taking a look at this issue, but I'm not sure how to test my fix. Any clues? :)

luigix25:
> @luigix25 @zhangjaycee any update on this?

Not really, I'm not working on this anymore.

germag:
should we close this PR then, and open an issue instead?

stefano-garzarella:
> should we close this PR then, and open an issue instead?

done

PR #95, opened 2021-10-21
build: Relax vm-memory dependency

rbradford:
By relaxing the dependency to ">=0.6" this crate can be used in projects
using newer vm-memory without pulling in duplicates of the crate.

Signed-off-by: Rob Bradford <robert.bradford@intel.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,9 @@ bitflags = ">=1.0.1"
 libc = ">=0.2.39"
 
 vmm-sys-util = ">=0.3.1"
-vm-memory = "0.6"
+vm-memory = ">=0.6"
 
 [dev-dependencies]
 tempfile = ">=3.2.0"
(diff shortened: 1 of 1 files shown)


sboeuf (approved):

Cargo.toml
```diff
@@ -24,9 +24,9 @@ bitflags = ">=1.0.1"
 libc = ">=0.2.39"
 
 vmm-sys-util = ">=0.3.1"
-vm-memory = "0.6"
+vm-memory = ">=0.6"
```
jiangliu, line 27: `vm-memory = "0.6"` should allow to upgrade to any version " >= 0.6 && < 1.0". Why does it cause multi-versions of vm-memory being used?

jiangliu: Does "cargo update" help to resolve the multi-version dependent crates issue?

jiangliu (approved):

PR #107, opened 2022-03-11
vhost_kern/vdpa: fix get_iova_range()

### LINKED ISSUE #106: vdpa: Incorrect implementation of get_iova_range()
Hi @stefano-garzarella 
I think there's an error with the current implementation of `get_iova_range` as it uses the wrong ioctl: https://github.com/rust-vmm/vhost/blob/main/src/vhost_kern/vdpa.rs#L142

Could you confirm if this is an issue?

stefano-garzarella:
### Summary of the PR

A wrong ioctl request was used in get_iova_range().
Unfortunately, there was no test, otherwise the ioctl would fail.

Let's fix this by using the right ioctl request.
We also add a test by checking the values we expect from
a vDPA-block simulator that returns [0, u64::MAX] range.

Fixes  #106

Reported-by: Sebastien Boeuf <sebastien.boeuf@intel.com>
Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/vhost_kern/vdpa.rs
+++ b/src/vhost_kern/vdpa.rs
@@ -139,7 +139,7 @@ impl<AS: GuestAddressSpace> VhostVdpa for VhostKernVdpa<AS> {
         let mut low_iova_range = vhost_vdpa_iova_range { first: 0, last: 0 };
 
         let ret =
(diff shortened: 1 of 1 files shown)


sboeuf (approved):

sboeuf:
Thanks for the quick fix @stefano-garzarella !

jiangliu (approved):

PR #112, opened 2022-07-12
src/lib.rs: add reason string to InvalidOperation

stsquad:
There are a number of reasons that a given vhost action may be
rejected as an InvalidOperation. Currently the returned error type
gives little clue resulting in user visible messages like:

  Error running daemon: HandleRequest(InvalidOperation)

By extending the error type with a static reason string we could
provide nicer feedback.

Signed-off-by: Alex Bennée <alex.bennee@linaro.org>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,7 +55,7 @@ pub mod vsock;
 #[derive(Debug)]
 pub enum Error {
     /// Invalid operations.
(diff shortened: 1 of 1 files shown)


stsquad:
I'm going to try a slightly different approach so I'll generate a new PR for it.

PR #120, opened 2022-10-06
vhost_kern: vdpa: Add missing ioctls

sboeuf:
New ioctls have been recently introduced to interact with vDPA devices.

This patch is based on Linux kernel v6.0, adding the list of missing ioctls:

- VHOST_VDPA_GET_CONFIG_SIZE
- VHOST_VDPA_GET_VQS_COUNT
- VHOST_VDPA_GET_GROUP_NUM
- VHOST_VDPA_GET_AS_NUM
- VHOST_VDPA_GET_VRING_GROUP
- VHOST_VDPA_SET_GROUP_ASID
- VHOST_VDPA_SUSPEND

### DIFF
--- a/src/vhost_kern/vdpa.rs
+++ b/src/vhost_kern/vdpa.rs
@@ -12,7 +12,7 @@ use std::os::unix::io::{AsRawFd, RawFd};
 use vm_memory::GuestAddressSpace;
 use vmm_sys_util::eventfd::EventFd;
 use vmm_sys_util::fam::*;
-use vmm_sys_util::ioctl::{ioctl_with_mut_ref, ioctl_with_ptr, ioctl_with_ref};
+use vmm_sys_util::ioctl::{ioctl, ioctl_with_mut_ref, ioctl_with_ptr, ioctl_with_ref};
 
 use super::vhost_binding::*;
 use super::{ioctl_result, Error, Result, VhostKernBackend, VhostKernFeatures};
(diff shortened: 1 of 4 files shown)


sboeuf:
I need to add some these new ioctls to the unit tests.

rbradford:
> I think we can add the test anyway and skip them (like we did with the `unwrap_not_found` macro) if the `ioctl` fails with `ENOIOCTLCMD` that should be the return value in vhost-vdpa kernel module when the `ioctl` is not implemented. What do you think?

Exactly what I was going to suggest. Add tests, validate them running locally and mark them as `#[ignore = "See #<bug number about kernel issue>"]`

sboeuf:
Sounds good I'll extend the tests but they won't fail if they're not supported.

src/vhost_kern/vdpa.rs
```diff
@@ -253,6 +304,12 @@ mod tests {
         };
     }
 
+    fn kernel_version() -> f32 {
```
stefano-garzarella, line 307: I honestly always avoid assuming features based on kernel version, because downstream versions don't necessarily have the same features, but since this is just a test it might be okay.

sboeuf:
@stefano-garzarella I think I addressed all of your comments, PTAL

src/vhost_user/connection.rs
```diff
@@ -356,7 +356,7 @@ impl<R: Req> Endpoint<R> {
                     .take(n)
                     .map(|fd| {
                         // Safe because we have the ownership of `fd`.
-                        unsafe { File::from_raw_fd(*fd) }
+                        File::from_raw_fd(*fd)
```
stefano-garzarella, line 359: Unrelated. Maybe better another commit or PR.

stefano-garzarella:
> @stefano-garzarella I think I addressed all of your comments, PTAL

Yep, thanks for that! I tested on a 5.16 kernel, and it worked well (skipping the unsupported IOCTLs.

I just left a small comment for an unrelated change that I think you accidentally added during a rebase (I was seeing the same warning yesterday as well)

stefano-garzarella (changes requested):

src/vhost_user/connection.rs
sboeuf: Ah yes I'll create a separate commit.

stefano-garzarella (approved):

sboeuf:
@slp @jiangliu PTAL

rbradford (approved):

PR #126, opened 2022-11-15
{vhost, vhost-user-backend}: Bulk update rust-vmm dependencies

rbradford:
Signed-off-by: Rob Bradford <robert.bradford@intel.com>

### DIFF
--- a/crates/vhost-user-backend/Cargo.toml
+++ b/crates/vhost-user-backend/Cargo.toml
@@ -11,13 +11,13 @@ license = "Apache-2.0"
 libc = "0.2.39"
 log = "0.4.17"
 vhost = { path = "../vhost", version = "0.5", features = ["vhost-user-slave"] }
-virtio-bindings = "0.1"
-virtio-queue = "0.6"
-vm-memory = { version = "0.9", features = ["backend-mmap", "backend-atomic"] }
-vmm-sys-util = "0.10"
+virtio-bindings = "0.1.0"
+virtio-queue = "0.7.0"
+vm-memory = { version = "0.10.0", features = ["backend-mmap", "backend-atomic"] }
+vmm-sys-util = "0.11.0"
 
 [dev-dependencies]
 nix = "0.25"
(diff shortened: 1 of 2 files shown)


stefano-garzarella (dismissed):

crates/vhost-user-backend/Cargo.toml
```diff
@@ -8,16 +8,16 @@ edition = "2018"
 license = "Apache-2.0"
 
 [dependencies]
-libc = "0.2.39"
+libc = "0.2.137"
```
slp, line 11: Does `vhost-user-backend` really need the latest version of `libc`?

Given that SemVer allows a dependency specified with caret version to be satisfied by the given version or any later version within the API context (for zero versions, any version with the same minor number), I think we should play conservative here. Otherwise, we could be making the distros packaging work a lot harder and imposing an arbitrary version to the consumers of the crate.

crates/vhost/Cargo.toml
```diff
@@ -24,13 +24,13 @@ vhost-user-master = ["vhost-user"]
 vhost-user-slave = ["vhost-user"]
 
 [dependencies]
-bitflags = "1.0"
-libc = "0.2.39"
+bitflags = "1.3.2"
```
slp, line 27: Same here. For non-zero versions, SemVer allows the `bitflags` dependency by any version that's equal or bigger than `1.0` and smaller than `2.0`.

slp (changes requested):

crates/vhost-user-backend/Cargo.toml
rbradford: Good plan. I updated the PR so it only touches the rust-vmm dependencies.

crates/vhost/Cargo.toml
rbradford: Ditto, this isn't in the updated version.

stefano-garzarella (approved):

slp (approved):

PR #135, opened 2022-12-21
vhost_user: fix unsoundness in VhostUserMsgHeader

alyssais:
### Summary of the PR

Since `VhostUserMsgHeader` implements `ByteValued`, it is supposed to be safe to construct from any correctly-sized arbitrary byte array. But that means we can do this:

```rust
let bytes = b"\xFF\xFF\xFF\xFF\x00\x00\x00\x00\x00\x00\x00\x00";
let header = VhostUserMsgHeader::<MasterReq>::from_slice(bytes).unwrap();
header.get_code()
```

constructing an invalid `MasterReq`, using only functions that are marked as safe.  Constructing an invalid enum value is undefined behavior in Rust, so this API is unsound.  This wasn't considered by the safety comment in `VhostUserMsgHeader::get_code`, which only considered the safety of requests that were valid enum variants.

If the vhost-user frontend process sends a message that the backend doesn't recognise, that's exactly what will happen, so the UB can be triggered from an external process (but a trusted one).

To fix this, we need to check whether the value is valid _before_ converting it.  `Req::is_valid` is changed to be a non-instance method, so it can be called before constructing the `Req`.
`VhostUserMsgHeader::get_code` is changed to return a `Result`, to accomodate the case where the request number is not a valid value for `R`.

Since this unsoundness is exposed to other processes, I will apply for an informational (not vulnerability) RUSTSEC advisory for this issue.

Signed-off-by: Alyssa Ross <alyssa.ross@unikie.com>

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/vhost/src/vhost_user/message.rs
+++ b/crates/vhost/src/vhost_user/message.rs
@@ -624,7 +624,7 @@ impl VhostUserVringAddr {
     }
 
     /// Create a new instance from `VringConfigData`.
-    #[cfg_attr(feature = "cargo-clippy", allow(clippy::identity_conversion))]
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::useless_conversion))]
     pub fn from_config_data(index: u32, config_data: &VringConfigData) -> Self {
         let log_addr = config_data.log_addr.unwrap_or(0);
         VhostUserVringAddr {
--- a/crates/vhost/src/vhost_user/message.rs
+++ b/crates/vhost/src/vhost_user/message.rs
@@ -14,6 +14,7 @@ use std::marker::PhantomData;
 
 use vm_memory::ByteValued;
 
+use super::{Error, Result};
 use crate::VringConfigData;
 
 /// The vhost-user specification uses a field of u32 to store message length.
@@ -45,7 +46,7 @@ pub const VHOST_USER_MAX_VRINGS: u64 = 0x8000u64;
 pub(super) trait Req:
     Clone + Copy + Debug + PartialEq + Eq + PartialOrd + Ord + Send + Sync + Into<u32>
 {
-    fn is_valid(&self) -> bool;
+    fn is_valid(value: u32) -> bool;
 }
 
 /// Type of requests sending from masters to slaves.
@@ -150,8 +151,8 @@ impl From<MasterReq> for u32 {
 }
 
 impl Req for MasterReq {
-    fn is_valid(&self) -> bool {
-        (*self > MasterReq::NOOP) && (*self < MasterReq::MAX_CMD)
+    fn is_valid(value: u32) -> bool {
+        (value > MasterReq::NOOP as u32) && (value < MasterReq::MAX_CMD as u32)
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/vhost/src/vhost_user/message.rs
```diff
@@ -981,38 +990,32 @@ mod tests {
...
     #[test]
     fn check_slave_request_code() {
-        let code = SlaveReq::NOOP;
-        assert!(!code.is_valid());
-        let code = SlaveReq::MAX_CMD;
-        assert!(!code.is_valid());
-        assert!(code > SlaveReq::NOOP);
+        assert!(!MasterReq::is_valid(MasterReq::NOOP as _));
+        assert!(!MasterReq::is_valid(MasterReq::MAX_CMD as _));
+        assert!(MasterReq::MAX_CMD > MasterReq::NOOP);
```
stefano-garzarella, line 1006: In this code block, we should use `SlaveReq::` methods/values

crates/vhost/src/vhost_user/slave_req_handler.rs
```diff
@@ -609,11 +609,10 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
...
             return Err(Error::InvalidMessage);
         }
-        let flags: VhostUserConfigFlags;
-        match VhostUserConfigFlags::from_bits(msg.flags) {
-            Some(val) => flags = val,
+        let flags: VhostUserConfigFlags = match VhostUserConfigFlags::from_bits(msg.flags) {
+            Some(val) => val,
             None => return Err(Error::InvalidMessage),
-        }
+        };
```
stefano-garzarella, line 615: This change seems unrelated (I prefer your cleanup), maybe you can do it in a separate commit.

stefano-garzarella (changes requested):
I left a few small comments, but overall it looks great!

crates/vhost/src/vhost_user/slave_req_handler.rs
lauralt: I think you can use `ok_or` here.

alyssais: I changed it because I saw CI failed due to clippy, but I think that may actually have been for a different reason, so I can probably drop this / make a separate PR to make clippy happy.

stefano-garzarella: Sorry, I misspoke!
The fix is fine, especially if there is a new clippy warning (I guess with a new version of Rust), I was just asking to do it in a separate commit, even within this PR :-)

crates/vhost/src/vhost_user/message.rs
alyssais: Fixed.

crates/vhost/src/vhost_user/slave_req_handler.rs
alyssais: Done. :)

stefano-garzarella (approved):

sboeuf (approved):

PR #142, opened 2023-02-23
vhost_user: Slave requests aren't only FS specific

vireshk:
There are a lot of slave requests supported by vhost-user protocol, it isn't just about FS_MAP and FS_UNMAP. Rename the files and declarations to slave request specific names.

### DIFF
--- a/crates/vhost-user-backend/src/backend.rs
+++ b/crates/vhost-user-backend/src/backend.rs
@@ -23,7 +23,7 @@ use std::ops::Deref;
 use std::sync::{Arc, Mutex, RwLock};
 
 use vhost::vhost_user::message::VhostUserProtocolFeatures;
-use vhost::vhost_user::SlaveFsCacheReq;
+use vhost::vhost_user::Slave;
 use vm_memory::bitmap::Bitmap;
 use vmm_sys_util::epoll::EventSet;
 use vmm_sys_util::eventfd::EventFd;
@@ -80,9 +80,7 @@ where
     ///
     /// A default implementation is provided as we cannot expect all backends to implement this
     /// function.
-    ///
-    /// TODO: this interface is designed only for vhost-user-fs, it should be refined.
-    fn set_slave_req_fd(&self, _vu_req: SlaveFsCacheReq) {}
+    fn set_slave_req_fd(&self, _slave: Slave) {}
 
     /// Get the map to map queue index to worker thread index.
     ///
@@ -163,9 +161,7 @@ where
     ///
     /// A default implementation is provided as we cannot expect all backends to implement this
     /// function.
-    ///
-    /// TODO: this interface is designed only for vhost-user-fs, it should be refined.
-    fn set_slave_req_fd(&mut self, _vu_req: SlaveFsCacheReq) {}
+    fn set_slave_req_fd(&mut self, _slave: Slave) {}
 
     /// Get the map to map queue index to worker thread index.
     ///
@@ -240,8 +236,8 @@ where
         self.deref().update_memory(mem)
     }
 
-    fn set_slave_req_fd(&self, vu_req: SlaveFsCacheReq) {
-        self.deref().set_slave_req_fd(vu_req)
+    fn set_slave_req_fd(&self, slave: Slave) {
+        self.deref().set_slave_req_fd(slave)
     }
 
     fn queues_per_thread(&self) -> Vec<u64> {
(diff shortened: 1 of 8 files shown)


crates/vhost/src/vhost_user/slave_req.rs
```diff
@@ -65,44 +67,44 @@ impl SlaveFsCacheReqInternal {
...
 #[derive(Clone)]
-pub struct SlaveFsCacheReq {
+pub struct Slave {
     // underlying Unix domain socket for communication
-    node: Arc<Mutex<SlaveFsCacheReqInternal>>,
+    node: Arc<Mutex<SlaveInternal>>,
 }
 
-impl SlaveFsCacheReq {
+impl Slave {
```
stefano-garzarella, line 85: Is `Slave` a bit generic?

Maybe `SlaveProxy` or `SlaveReqProxy` (I'm not good with names :-)

(i'd like one day to rename Master/Slave to frontend/backend, but I don't think it is for this PR)

crates/vhost/src/vhost_user/slave_req.rs
```diff
@@ -126,7 +128,7 @@ impl SlaveFsCacheReq {
     }
 }
 
-impl VhostUserMasterReqHandler for SlaveFsCacheReq {
+impl VhostUserMasterReqHandler for Slave {
```
stefano-garzarella, line 131: Since we are here, should we factor out `fs_*` methods from the `VhostUserMasterReqHandler` trait?

crates/vhost-user-backend/src/backend.rs
```diff
@@ -82,7 +82,7 @@ where
     /// function.
     ///
     /// TODO: this interface is designed only for vhost-user-fs, it should be refined.
```
stefano-garzarella, line 84: Should we remove this TODO?

stefano-garzarella: And also on line 85 and 168

crates/vhost/src/vhost_user/slave_req.rs
vireshk: This is how `Slave` and `Master` look after this change:

`
pub struct Slave {
    // underlying Unix domain socket for communication
    node: Arc<Mutex<SlaveInternal>>,
}

pub struct Master {
    node: Arc<Mutex<MasterInternal>>,
}
`

And so I went with just `Slave` here.

crates/vhost/src/vhost_user/slave_req.rs
vireshk: This is how `Slave` and `Master` look after this change:

```
pub struct Slave {
    // underlying Unix domain socket for communication
    node: Arc<Mutex<SlaveInternal>>,
}

pub struct Master {
    node: Arc<Mutex<MasterInternal>>,
}
```

And so I went with just `Slave` here.

crates/vhost/src/vhost_user/slave_req.rs
vireshk: I am not sure why we should do that.

`VhostUserSlaveReqHandler` trait defines the helpers required by the backend to process messages from frontend.

Similarly, `VhostUserMasterReqHandler` trait defines the helpers required by the frontend to process messages from backend. And the `FS_*` messages should be handled from within `VhostUserMasterReqHandler`.

Is my understanding correct ?

crates/vhost-user-backend/src/backend.rs
vireshk: Done.

crates/vhost/src/vhost_user/slave_req.rs
stefano-garzarella: I'm confused, I interpreted that `fs_*` functions should be implemented only by backends that implement a `virtio-fs` device (e.g. virtiofsd)

crates/vhost/src/vhost_user/slave_req.rs
stefano-garzarella: Got it, good for now!

crates/vhost/src/vhost_user/slave_req.rs
vireshk: Yes that is correct, but they still must be part of VhostUserMasterReqHandler, that is where all requests from backend are collected. Many of the requests are optional and are only required to be implemented by the specific use cases, like virtiofsd in this case.

stefano-garzarella: Okay, got it!

stefano-garzarella (approved):

jiangliu (approved):

PR #153, opened 2023-04-25
Fix set_vring_addr issues

justxuewei:
`VhostBackend::set_vring_addr()` receives a vring config data which
contains the addresses of desc table, used ring and avail ring.
`VhostBackend::is_valid()` checks the addresses in the guest address space.
`VHOST_SET_VRING_ADDR` uses the addresses in the host address space.
However, the method doesn't convert those addresses.

To address this issue, the addresses passed by the config are checked in
the guest address space. Then, they are converted by
`VringConfigData::to_vhost_vring_addr()` into the host address space to
setup the vring on the kernel.

Signed-off-by: Xuewei Niu <niuxuewei.nxw@antgroup.com>

### DIFF
--- a/crates/vhost/src/vhost_kern/mod.rs
+++ b/crates/vhost/src/vhost_kern/mod.rs
@@ -200,20 +200,15 @@ impl<T: VhostKernBackend> VhostBackend for T {
     ///
     /// # Arguments
     /// * `queue_index` - Index of the queue to set addresses for.
-    /// * `config_data` - Vring config data.
+    /// * `config_data` - Vring config data, addresses of desc_table, avail_ring
+    ///     and used_ring are in the guest address space.
     fn set_vring_addr(&self, queue_index: usize, config_data: &VringConfigData) -> Result<()> {
         if !self.is_valid(config_data) {
             return Err(Error::InvalidQueue);
         }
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
 /build
 /kcov_build
 /target
+.vscode
```
ZizhengBian, line 4: It's better to split this line into another patch.

justxuewei: Fixed

stefano-garzarella (dismissed):
LGTM!

Maybe we should add a test to cover this scenario.

eryugey (approved):

crates/vhost/src/backend.rs
```diff
@@ -57,6 +61,35 @@ impl VringConfigData {
             0
         }
     }
+
+    #[cfg(feature = "vhost-kern")]
```
jiangliu, line 65: Can we move this changes into `src/vhost_kern/mod.rs`, so we can avoid those condition compilations.

justxuewei: I moved the method into `src/vhost_kern/mod.rs`, Please take a look, thanks.

stefano-garzarella (approved):

jiangliu (approved):

PR #154, opened 2023-05-04
 Fix return value of `GET_VRING_BASE` message

germag:
### Summary of the PR

 When the frontend sends the `GET_VRING_BASE` message, we should return the vring's last available index and stop the vring.  Commit https://github.com/rust-vmm/vhost/commit/17131359ebae207476d009a65b702f028b581622 reset the vring upon receiving `GET_VRING_BASE`, But to return the correct index we should not reset the queue before getting its value, otherwise we will always return 0.

Note:
Neither qemu[0] nor dpdk[1] reset the vring when receiving `GET_VRING_BASE`

[0] https://github.com/qemu/qemu/blob/792f77f376adef944f9a03e601f6ad90c2f891b2/subprojects/libvhost-user/libvhost-user.c#L1175

[1] https://github.com/DPDK/dpdk/blob/d03446724972d2a1bb645ce7f3e64f5ef0203d61/lib/vhost/vhost_user.c#L2133

### DIFF
--- a/crates/vhost-user-backend/src/handler.rs
+++ b/crates/vhost-user-backend/src/handler.rs
@@ -386,6 +386,8 @@ where
             }
         }
 
+        let next_avail = self.vrings[index as usize].queue_next_avail();
+
         self.vrings[index as usize].set_kick(None);
         self.vrings[index as usize].set_call(None);
 
@@ -396,8 +398,6 @@ where
             .get_mut()
             .get_queue_mut()
             .reset();
-
-        let next_avail = self.vrings[index as usize].queue_next_avail();
 
         Ok(VhostUserVringState::new(index, u32::from(next_avail)))
     }


stefano-garzarella (commented):
LGTM, but  I wonder whether we should now call reset() for example in set_owner() or reset_owner()

stefano-garzarella:
> According to the spec `RESET_OWNER` is deprecated:

I see, thanks for checking!

> 
> Not sure about doing in `SET_OWNER` qemu already set a bunch of stuff before sending `SET_OWNER`, same for the first kick (mentioned in the reset comment). The [spec](https://qemu-project.gitlab.io/qemu/interop/vhost-user.html#ring-states) says we should _stop_ the vring, but not resetting it, and the vring is initialized when we add the memory regions, so not sure if we should keep the `reset()`, btw neither qemu[0] nor dpdk[1] reset the vring when receiving `GET_VRING_BASE`, @slp wdyt?

Okay, the current version leaves `reset()` and fetches the next_avail before it. This LGTM, but I think your first version (avoid to call `reset()` and just call `set_ready(false)`) was okay, since it seems libvhost-user does the same.

Anyway, I'd wait @slp since that change was part of https://github.com/rust-vmm/vhost-user-backend/pull/18

germag:
(Sorry, I forgot to comment the diff between v1 and v2)

v2:
- Get the last available index before changing the vring status
- Keep the call to `reset()` instead of  `set_ready(false)`

jiangliu (approved):

jiangliu (approved):

jiangliu (approved):

jiangliu (approved):

stefano-garzarella:
@jiangliu are you okay with resetting the vring?

As pointed out by German, neither qemu's libvhost-user nor dpdk reset the vring when receiving GET_VRING_BASE.

jiangliu:
> @jiangliu are you okay with resetting the vring?
> 
> As pointed out by German, neither qemu's libvhost-user nor dpdk reset the vring when receiving GET_VRING_BASE.

It would be safer to keep the same behavior as qemu/libvhost-user:)

stefano-garzarella:
> Current version LGTM.

It's not clear to me though why we have to reset the whole VQ, would it be better just to stop it?

stefano-garzarella:
So, let's merge this version that should have less impact, then we can try to figure out if we can avoid resetting the VQ and if we need to move that call in some other place

stefano-garzarella:
> > > Current version LGTM.
> > 
> > 
> > It's not clear to me though why we have to reset the whole VQ, would it be better just to stop it?
> 
> I'm honestly not sure, but I'd prefer to cut a release without altering that behavior, and then consider a different approach (along with things like #155) in the next one. Sounds reasonable?

Yep, same conclusion https://github.com/rust-vmm/vhost/pull/154#issuecomment-1551175318 :-)

PR #155, opened 2023-05-10
vhost-user-backend: simplify the use of generics

Ablu:
In order to allow zero-cost exchanging of the concrete bitmap and vring
types, a lot of the generic code required using a tuple of
`<VringType, BitmapType>` for parameterizing the impl's. Once code is
also oblivious of the concrete backend (and a lot of code is), this
tuple turns into a triplet. Juggling these three single letter generic
parameters while making sure that all the type constraints (that are
also changing depending on the abstraction layer) does not feel very
ergonomic.

While one can argue that within this crate, this would be fine since
people probably know the internals, the design choice is leaking out
into every consumer of vhost-user-backend. Instead of just being able
to reference "some backend" one needs to copy a lot of boilerplate code
for also passing the other type parameters (again, while making sure
that all the constraints are met).

Instead, this commit changes things to utilize associated types [1].
This makes the Bitmap and Vring types part of the backend trait and
requires the implementations to spell them out. Code that just wants to
use the backend without needing to know the details can now just use
the trait without needing to specify the Bitmap and Vring types again.
Where needed, restricting Bitmap and Vring further is still possible
(though one no longer needs to copy all the existing restrictions and
can keep the code more maintainable by only listing new ones).

Overall, my main target was to improve the ergonomics of the consumers
of the crate. But I think the change also improves the readability and
maintainability within this crate. Combined, this hopefully justifies
the small code breakage in consumers.

No functional changes intended.
No change in type flexibility intended.

BREAKING CHANGE, consumers of the lib will need to adjust their code
(but it should improve the general readability).

Signed-off-by: Erik Schilling <erik.schilling@linaro.org>

[1] https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#specifying-placeholder-types-in-trait-definitions-with-associated-types

### Summary of the PR

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/vhost-user-backend/CHANGELOG.md
+++ b/crates/vhost-user-backend/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### Changed
 - Change uses of master/slave for frontend/backend in the codebase.
 - [[#192]](https://github.com/rust-vmm/vhost/pull/192) vhost-user-backend: remove return value from handle_event
+- [[#155]](https://github.com/rust-vmm/vhost/pull/155) Converted generic type
+  parameters of VhostUserBackend into associated types.
 
 ### Fixed
 
--- a/crates/vhost-user-backend/src/lib.rs
+++ b/crates/vhost-user-backend/src/lib.rs
@@ -14,7 +14,6 @@ use std::sync::{Arc, Mutex};
 use std::thread;
 
 use vhost::vhost_user::{BackendListener, BackendReqHandler, Error as VhostUserError, Listener};
-use vm_memory::bitmap::Bitmap;
 use vm_memory::mmap::NewBitmap;
 use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
 
@@ -81,17 +80,17 @@ pub type Result<T> = std::result::Result<T, Error>;
 ///
 /// This structure is the public API the backend is allowed to interact with in order to run
 /// a fully functional vhost-user daemon.
-pub struct VhostUserDaemon<S, V, B: Bitmap + 'static = ()> {
+pub struct VhostUserDaemon<T: VhostUserBackend> {
     name: String,
-    handler: Arc<Mutex<VhostUserHandler<S, V, B>>>,
+    handler: Arc<Mutex<VhostUserHandler<T>>>,
     main_thread: Option<thread::JoinHandle<Result<()>>>,
 }
 
-impl<S, V, B> VhostUserDaemon<S, V, B>
+impl<T> VhostUserDaemon<T>
 where
-    S: VhostUserBackend<V, B> + Clone + 'static,
-    V: VringT<GM<B>> + Clone + Send + Sync + 'static,
-    B: NewBitmap + Clone + Send + Sync,
+    T: VhostUserBackend + Clone + 'static,
+    T::Bitmap: NewBitmap + Clone + Send + Sync,
+    T::Vring: Clone + Send + Sync,
 {
     /// Create the daemon instance, providing the backend implementation of `VhostUserBackend`.
     ///
@@ -100,8 +99,8 @@ where
     /// but they get to be registered later during the sequence.
     pub fn new(
         name: String,
-        backend: S,
-        atomic_mem: GuestMemoryAtomic<GuestMemoryMmap<B>>,
+        backend: T,
+        atomic_mem: GuestMemoryAtomic<GuestMemoryMmap<T::Bitmap>>,
     ) -> Result<Self> {
         let handler = Arc::new(Mutex::new(
             VhostUserHandler::new(backend, atomic_mem).map_err(Error::NewVhostUserHandler)?,
@@ -122,7 +121,7 @@ where
     /// it acts as a client or a server.
     fn start_daemon(
         &mut self,
-        mut handler: BackendReqHandler<Mutex<VhostUserHandler<S, V, B>>>,
+        mut handler: BackendReqHandler<Mutex<VhostUserHandler<T>>>,
     ) -> Result<()> {
         let handle = thread::Builder::new()
             .name(self.name.clone())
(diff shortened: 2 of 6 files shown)


Ablu:
CI seems to have timed out? Probably only needs a retrigger...

jiangliu:
Once upon a time, we have struggled with rust generic associated types when developing the vm-memory crate, not sure about the latest progess about support of GAT. One good news is that vhost-user-backend is not as `fundamental` as vm-memory, so GAT may not be an issue for vhost-user-backend.

Ablu:
@jiangliu hm, do you recall the concrete problems? Maybe it was with older rust versions? The only catch that I can think of is that you cannot use a normal generic type parameter in an associated type (I think that only works with dynamic dispatch). But that is a pretty weird case anyway (why not simply leave the type open then and set it on a higher level)?

jiangliu (approved):

stsquad:
Do you have any examples of the conversion needed to show how its an improvement?

Ablu:
@stsquad: The changes done to `VhostUserHandler` are probably a good hint at how things improve for consumers. Essentially, the number of generic arguments is simply reduced which makes them a lot saner to write. Right now, every time you want to pass a backend around, you will have to do so while juggling all these parameters (and copying the constraints all over the place).

I can draft a proposal how I was planning the simplify vhost-user in the next days.

Ablu:
@stsquad here is an example train of thought for vhost-user:
1. Assume that we want to move the boilerplate around creating the daemon into a separate function (for example, in order to share it with the other daemons, or move it into some shared lib)
2. The code for that, right now, could look like: https://github.com/rust-vmm/vhost-device/commit/a447db5840ba47a0051400814b49dfda508990eb
3. Forcing the user to spell out, for example, the full VringT type does not looks particulary helpful
4. https://github.com/Ablu/vhost-device/commit/ed5cafbaa2a7cb72f5ea0db4aa295324f4ce0294 now show the changes necessary / possible with the associated type changes of this PR.
5. Mostly, one does not need to spell out ALL the requirements. One only needs to list what `VhostUserDaemon` puts on top of the `VhostUserBackend`. I also feel the associated types a bit easier to reason, given that they a) have more descriptive names, b) are immediately understood as types of the backend. Of course there may still be further room for improvement here, but I argue that the new code requires quite a few brain cycles less to process.

The benefits may be larger if code is generic over a VhostUserBackend, but does not need to fulfill further requirements (unlike what `VhostUserDaemon` demands here). In those cases the entire `where` could would probably become superfluous. However, I did not find code that uses things in vhost-device or virtiofsd.

Let me know if anything is unclear :).

Ablu:
https://github.com/rust-vmm/vhost-device/pull/362 would be another example of code where this could simplify things.

Ablu:
rebased

stefano-garzarella:
@Ablu Sorry to just get here now, but I really like this cleanup!
It's going to be quite a change for our users, but I think it's worth it!


Would you like to rebase?
I hope this is the right release cycle to include this change.

Ablu:
rebased

sboeuf (approved):

crates/vhost-user-backend/src/lib.rs
```diff
@@ -12,9 +12,7 @@ use std::fmt::{Display, Formatter};
 use std::path::Path;
 use std::sync::{Arc, Mutex};
 use std::thread;
-
```
stefano-garzarella, line 15: Maybe accidentally happened during the rebase, but we enforce to keep that groups separated.

crates/vhost-user-backend/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Changed
 - Change uses of master/slave for frontend/backend in the codebase.
+- BREAKING: Converted generic type parameters of VhostUserBackend into
```
stefano-garzarella, line 9: We usually put the link to the PR with the title, but I don't know why we changed with the other merged PR.

I'd rather still have the link, but we can add it later when we prepare the release.

stefano-garzarella (commented):
LGTM, just some minor comments.

crates/vhost-user-backend/CHANGELOG.md
stsquad: I guess this also means a major version jump?

Ablu: > I guess this also means a major version jump?

Yes, but the other change - which renamed types - already requires that anyway. Thats why it is a good idea to batch these two changes together :)

stefano-garzarella: > I guess this also means a major version jump?

Yep! Since we still are on `0.y.z`, increasing `y` is a major version jump: https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility

Ablu: > We usually put the link to the PR with the title, but I don't know why we changed with the other merged PR.
> 
> I'd rather still have the link, but we can add it later when we prepare the release.

added. Also dropped the `BREAKING:` since the other change - which also is breaking in a similar way - does not have it.

stefano-garzarella (approved):

Ablu:
> @Ablu do you plan to work on https://github.com/rust-vmm/vhost-device Draft PR to be merged when we do the next release of this crate?

Will prepare one!

Ablu:
> Will prepare one!

https://github.com/rust-vmm/vhost-device/pull/461

stsquad (approved):

epilys (approved):

stefano-garzarella:
@jiangliu @sboeuf you already approved a previous version, this one has only a few minimal changes, can you take a look again? Thanks ;-)

sboeuf (approved):

PR #160, opened 2023-05-19
Xen/mmapv2

vireshk:
vhost: Add support for generic memory mappings
    
vm-memory crate now supports Xen specific memory mappings via a special
feature: "xen". Modify all vhost crates to work with the new feature.
    
Signed-off-by: Viresh Kumar <viresh.kumar@linaro.org>

### DIFF
--- a/crates/vhost-user-backend/CHANGELOG.md
+++ b/crates/vhost-user-backend/CHANGELOG.md
@@ -9,6 +9,14 @@
 
 ### Deprecated
 
+## v0.10.0
+
+### Added
+- [[#169]](https://github.com/rust-vmm/vhost/pull/160) vhost-user-backend: Add support for Xen memory mappings
+
+### Fixed
+- [[#161]](https://github.com/rust-vmm/vhost/pull/161) get_vring_base should not reset the queue
+
 ## v0.9.0
 
 ### Added
--- a/crates/vhost-user-backend/src/handler.rs
+++ b/crates/vhost-user-backend/src/handler.rs
@@ -22,7 +22,7 @@ use virtio_bindings::bindings::virtio_ring::VIRTIO_RING_F_EVENT_IDX;
 use virtio_queue::{Error as VirtQueError, QueueT};
 use vm_memory::bitmap::Bitmap;
 use vm_memory::mmap::NewBitmap;
-use vm_memory::{FileOffset, GuestAddress, GuestAddressSpace, GuestMemoryMmap, GuestRegionMmap};
+use vm_memory::{GuestAddress, GuestAddressSpace, GuestMemoryMmap, GuestRegionMmap};
 use vmm_sys_util::epoll::EventSet;
 
 use super::backend::VhostUserBackend;
@@ -277,15 +277,19 @@ where
     ) -> VhostUserResult<()> {
         // We need to create tuple of ranges from the list of VhostUserMemoryRegion
         // that we get from the caller.
-        let mut regions: Vec<(GuestAddress, usize, Option<FileOffset>)> = Vec::new();
+        let mut regions = Vec::new();
         let mut mappings: Vec<AddrMapping> = Vec::new();
 
         for (region, file) in ctx.iter().zip(files) {
-            let g_addr = GuestAddress(region.guest_phys_addr);
-            let len = region.memory_size as usize;
-            let f_off = FileOffset::new(file, region.mmap_offset);
-
-            regions.push((g_addr, len, Some(f_off)));
+            regions.push(
+                GuestRegionMmap::new(
+                    region.mmap_region(file)?,
+                    GuestAddress(region.guest_phys_addr),
+                )
+                .map_err(|e| {
+                    VhostUserError::ReqHandlerError(io::Error::new(io::ErrorKind::Other, e))
+                })?,
+            );
             mappings.push(AddrMapping {
                 vmm_addr: region.user_addr,
                 size: region.memory_size,
@@ -293,7 +297,7 @@ where
             });
         }
 
-        let mem = GuestMemoryMmap::from_ranges_with_files(regions).map_err(|e| {
+        let mem = GuestMemoryMmap::from_regions(regions).map_err(|e| {
             VhostUserError::ReqHandlerError(io::Error::new(io::ErrorKind::Other, e))
         })?;
 
@@ -526,10 +530,9 @@ where
         file: File,
     ) -> VhostUserResult<()> {
         let guest_region = Arc::new(
-            GuestRegionMmap::from_range(
+            GuestRegionMmap::new(
+                region.mmap_region(file)?,
                 GuestAddress(region.guest_phys_addr),
-                region.memory_size as usize,
-                Some(FileOffset::new(file, region.mmap_offset)),
             )
             .map_err(|e| {
                 VhostUserError::ReqHandlerError(io::Error::new(io::ErrorKind::Other, e))
--- a/crates/vhost/src/backend.rs
+++ b/crates/vhost/src/backend.rs
@@ -10,12 +10,15 @@
 //! Common traits and structs for vhost-kern and vhost-user backend drivers.
 
 use std::cell::RefCell;
+use std::os::unix::io::AsRawFd;
 use std::os::unix::io::RawFd;
 use std::sync::RwLock;
 
+use vm_memory::{bitmap::Bitmap, Address, GuestMemoryRegion, GuestRegionMmap};
 use vmm_sys_util::eventfd::EventFd;
 
-use super::Result;
+use super::vhost_user::message::{VhostUserMemoryRegion, VhostUserSingleMemoryRegion};
+use super::{Error, Result};
 
 /// Maximum number of memory regions supported.
 pub const VHOST_MAX_MEMORY_REGIONS: usize = 255;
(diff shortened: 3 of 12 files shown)


crates/vhost/src/vhost_user/message.rs
```diff
@@ -500,18 +544,74 @@ impl VhostUserMemoryRegion {
...
+    /// Creates range from Self.
+    pub fn as_range(&self, file: File) -> (GuestAddress, usize, Option<FileOffset>) {
+        let g_addr = GuestAddress(self.guest_phys_addr);
+        let len = self.memory_size as usize;
+        let f_off = FileOffset::new(file, self.mmap_offset);
+
+        (g_addr, len, Some(f_off))
+    }
+
+    /// Creates range from Self.
```
Ablu, line 557: This looks like a copy paste error? Overall, I currently have the feeling that most doc strings do not really add too much compared to what the function signature already gives away. Maybe some more detail could be helpful here and there...

Ablu (commented):
@vireshk I have not found anything that jumped to my eye. However, I find it a bit hard to follow what is happening. I think a bit more detail in the commit messages and a few more details in the doc strings could be helpful to follow what happens here :).

crates/vhost/src/vhost_user/message.rs
```diff
@@ -504,18 +541,74 @@ impl VhostUserMemoryRegion {
             mmap_offset,
         }
     }
+
+    /// Creates mmap region from Self.
+    pub fn as_mmap_region<B: NewBitmap>(&self, file: File) -> Result<MmapRegion<B>> {
```
Ablu, line 546: I would consider dropping the `as_` here

crates/vhost/src/vhost_user/message.rs
```diff
@@ -528,44 +529,36 @@ pub type VhostUserMemoryPayload = Vec<VhostUserMemoryRegion>;
...
-            || self.guest_phys_addr.checked_add(self.memory_size).is_none()
-            || self.user_addr.checked_add(self.memory_size).is_none()
-            || self.mmap_offset.checked_add(self.memory_size).is_none()
-        {
-            return false;
-        }
-        true
-    }
-}
+impl VhostUserMsgValidator for VhostUserSingleMemoryRegion {}
```
Ablu, line 680: This can be just dropped with the Deref impl in place, or am I missing something?

crates/vhost-user-backend/tests/vhost-user-server.rs
```diff
@@ -166,13 +166,13 @@ fn vhost_user_client(path: &Path, barrier: Arc<Barrier>) {
...
-        mmap_offset: 0,
-        mmap_handle: fd.file().as_raw_fd(),
-    }];
+    let regions = [VhostUserMemoryRegionInfo::new(
+        0x100000,
+        0x100000,
+        addr,
+        0,
+        fd.file().as_raw_fd(),
+    )];
```
Ablu, line 175: rust-analyzer provides the context as code lenses anyway... Might leave people guessing the context of the magic numbers a bit more when not using rust-analyzer, but I guess if it simplifies the code thats ok I guess.

crates/vhost/src/vhost_user/message.rs
vireshk: That gives:

```
the trait `message::VhostUserMsgValidator` is not implemented for `message::VhostUserSingleMemoryRegion``
```

crates/vhost-user-backend/tests/vhost-user-server.rs
vireshk: > rust-analyzer provides the context as code lenses anyway... Might leave people guessing the context of the magic numbers a bit more when not using rust-analyzer, but I guess if it simplifies the code thats ok I guess.

And this was also part of the cleanups series that got meged :)

vireshk:
Waiting for https://github.com/rust-vmm/vm-virtio/pull/250 to get merged.

crates/vhost/src/vhost_user/message.rs
Ablu: Hm... I did not see where it was used as type, but looks like I commented on unrelated stuff anyway.

crates/vhost/src/backend.rs
```diff
@@ -72,6 +75,71 @@ pub struct VhostUserMemoryRegionInfo {
...
+    pub xen_mmap_data: u32,
+}
+
+impl VhostUserMemoryRegionInfo {
+    /// Creates Self from GuestRegionMmap.
+    pub fn from_guest_region<B: Bitmap>(region: &GuestRegionMmap<B>) -> Result<Self> {
+        let (mmap_handle, mmap_offset) = match region.file_offset() {
+            Some(file_offset) => (file_offset.file().as_raw_fd(), file_offset.start()),
+            None => return Err(Error::InvalidGuestMemoryRegion),
+        };
```
Ablu, line 94: probably would have used `.ok_or[_else](...)?` here. But I do not think this code is complex either...

crates/vhost/src/vhost_user/message.rs
```diff
@@ -492,8 +506,31 @@ pub struct VhostUserMemoryRegion {
...
+    fn validate(&self) -> bool {
+        if self.memory_size == 0
+            || self.guest_phys_addr.checked_add(self.memory_size).is_none()
+            || self.user_addr.checked_add(self.memory_size).is_none()
+            || self.mmap_offset.checked_add(self.memory_size).is_none()
+        {
+            return false;
+        }
+
+        true
```
Ablu, line 529: just negate the if statement and drop the branching?

```rs
    fn validate(&self) -> bool {
        self.memory_size != 0
            && self.guest_phys_addr.checked_add(self.memory_size).is_some()
            && self.user_addr.checked_add(self.memory_size).is_some()
            && self.mmap_offset.checked_add(self.memory_size).is_some())
    }
```
Maybe one could also call this something like `is_map_info_valid` in order to prevent the confusion of having `is_valid` and `validate`?

Ablu (commented):
just some nit-picking...

vireshk:
Fixes: https://github.com/rust-vmm/vhost/issues/169

Ablu (approved):

crates/vhost/src/vhost_user/message.rs
```diff
@@ -492,8 +506,26 @@ pub struct VhostUserMemoryRegion {
...
+impl VhostUserMemoryRegion {
+    fn is_valid_common(&self) -> bool {
+        self.memory_size != 0
+            && self.guest_phys_addr.checked_add(self.memory_size).is_some()
+            && self.user_addr.checked_add(self.memory_size).is_some()
+            && self.mmap_offset.checked_add(self.memory_size).is_some()
+    }
 }
 
+#[cfg(not(feature = "xen"))]
```
stefano-garzarella, line 528: I'm a little concerned about this different implementation if `xen` is enabled or not.

Should we document that if the `xen` feature is enabled, then this crate doesn't work except with xen?

stefano-garzarella: Maybe we can introduce a trait with 2 implementations that can be used together, or this will never happen?

crates/vhost/src/vhost_user/slave_req_handler.rs
```diff
@@ -432,6 +432,9 @@ impl<S: VhostUserSlaveReqHandler> SlaveReqHandler<S> {
                 self.acked_protocol_features = msg.value;
                 self.update_reply_ack_flag();
                 self.send_ack_message(&hdr, res)?;
+
+                #[cfg(feature = "xen")]
+                self.check_proto_feature(VhostUserProtocolFeatures::XEN_MMAP)?;
```
stefano-garzarella, line 437: Okay, so here we always fail if `xen` feature is enabled, but `XEN_MAP` not negotiated. I think we should document this, but in general I'd prefer that the presence of a feature or not is handled at runtime if it's not too complicated.

crates/vhost-user-backend/src/handler.rs
```diff
@@ -441,7 +441,13 @@ where
     }
 
     fn get_protocol_features(&mut self) -> VhostUserResult<VhostUserProtocolFeatures> {
-        Ok(self.backend.protocol_features())
+        let features = self.backend.protocol_features();
+
+        // Enable the `XEN_MMAP` protocol feature for backends if xen feature is enabled.
+        #[cfg(feature = "xen")]
+        let features = features | VhostUserProtocolFeatures::XEN_MMAP;
```
stefano-garzarella, line 448: If we force the flag here, then we should also check that it has been negotiated correctly in `set_protocol_features()`.

stefano-garzarella (commented):
The thing that is not clear to me is: if we enable `xen` then we will always only handle xen guests, and we could never have the case of handling both?

If this is the case, then we should document this compile-time feature like: if you enable it you can only handle xen guests.

crates/vhost/src/vhost_user/message.rs
```diff
@@ -422,6 +427,15 @@ bitflags! {
...
+        /// Support Xen mmap.
+        const XEN_MMAP = 0x0002_0000;
+    }
+}
+
+impl VhostUserProtocolFeatures {
+    /// XEN_MMAP is set.
+    pub fn is_xen_mmap(features: u64) -> bool {
+        features & Self::XEN_MMAP.bits() != 0
     }
```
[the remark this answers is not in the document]
vireshk: not needed anymore.

vireshk:
> The thing that is not clear to me is: if we enable `xen` then we will always only handle xen guests, and we could never have the case of handling both?

This is based on the decision taken in the vm-memory crate. In the very first version, I tried to support both kvm and Xen mappings together in a single build, but that was rejected in favor of simplifying the code and selecting one at compile time only.

And so yes, this is either Xen or kvm, but not both currently. Though it may change later, but current code is selected at build time only. Will document this.

crates/vhost/src/vhost_user/message.rs
```diff
@@ -492,8 +506,26 @@ pub struct VhostUserMemoryRegion {
     pub user_addr: u64,
     /// Offset where region starts in the mapped memory.
     pub mmap_offset: u64,
+
+    #[cfg(feature = "xen")]
```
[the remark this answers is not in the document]
vireshk: > What happens if `xen` build feature is enabled but the `XEN_MAP` protocol feature is not negotiated at runtime?

We should fail then, as what the code does [here[(https://github.com/rust-vmm/vhost/pull/160/files#diff-45488e39bfb65df325f876039433282983d25c9040a9f8f646535b50eedb6426R437).

> Should we add a new `VhostUserMemoryRegionXen` to allow the user to use both structure at runtime, also if this crate is built with the `xen` feature?

No. We decided to make only one of these work at compile time in the vm-memory crate.

crates/vhost/src/vhost_user/message.rs
vireshk: Will not happen unless vm-memory is updated too.

crates/vhost/src/vhost_user/slave_req_handler.rs
vireshk: With `if cfg!()` ? Or something else ? Though since we used the compile time option everywhere in vm-memory crate, it made sense to do the same here since both can't be used at once.

crates/vhost-user-backend/src/handler.rs
vireshk: > If we force the flag here, then we should also check that it has been negotiated correctly in `set_protocol_features()`.

Yeah, something like [this](https://github.com/rust-vmm/vhost/pull/160/files#diff-45488e39bfb65df325f876039433282983d25c9040a9f8f646535b50eedb6426R437), right ?

vireshk:
@stefano-garzarella updated code, hopefully covered all your concerns apart from https://github.com/rust-vmm/vhost/pull/160#discussion_r1251601495.

crates/vhost/src/vhost_user/slave_req_handler.rs
stefano-garzarella: Yep, this is fine now I get how the xen support works ;-)

crates/vhost-user-backend/src/handler.rs
stefano-garzarella: It is strange to set it here and then check it in the other crate.

Can't we do everything here and check that it is set a few lines down in `VhostUserHandler::set_protocol_features()`?

Or I'm missing somenthig?

crates/vhost-user-backend/CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # Changelog
-## [Unreleased]
```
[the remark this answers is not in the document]
stefano-garzarella: We also leave the structure ready for the next cycle. Something like this:
```
## [Unreleased]

### Added

### Changed

### Fixed

### Deprecated

```

For example, you can checkout the `vhost-v0.7.0` tag to see it.

crates/vhost-user-backend/src/handler.rs
vireshk: Ah, makes sense. I have moved them both to the vhost crate as that is more central to everything.

crates/vhost-user-backend/CHANGELOG.md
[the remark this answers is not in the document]
vireshk: I  have tried to fix them all this time, fingers crossed  :)

vireshk:
Hmm, I have pushed my branch again: https://github.com/vireshk/vhost/tree/xen/mmapv2 , but somehow this pull request isn't fetching the changes automatically anymore. Any idea ? 

@stefano-garzarella @jiangliu

vireshk:
@dependabot rebase

vireshk:
@dependabot rebase

vireshk:
> Hmm, I have pushed my branch again: https://github.com/vireshk/vhost/tree/xen/mmapv2 , but somehow this pull request isn't fetching the changes automatically anymore. Any idea ?

Okay, it worked now :)

stefano-garzarella (approved):

jiangliu (approved):

PR #161, opened 2023-05-22
get_vring_base should not reset the queue

germag:
### Summary of the PR

The spec specifies that on receiving `GET_VRING_BASE` the backend should stop the vring, but not that it must be reset. This is intended for `VHOST_USER_RESET_DEVICE`, also in this case the spec makes a difference between stopping and disabling the ring.

The spec also doesn't forbid to send `VHOST_USER_SET_VRING_ENABLE` to enable the vring after receiving `GET_VRING_BASE` or sending more `GET_VRING_BASE` messages, which would always respond 0. Moreover, qemu[0] doesn't reset the vring either.

[0] https://github.com/qemu/qemu/blob/792f77f376adef944f9a03e601f6ad90c2f891b2/subprojects/libvhost-user/libvhost-user.c#L1175

Note:
I just added blank lines around `self.vrings[index as usize].set_queue_ready(false);` because the first time I worked on this (#154) I didn't realize that the ring was already being disabled.

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.

### DIFF
--- a/crates/vhost-user-backend/src/handler.rs
+++ b/crates/vhost-user-backend/src/handler.rs
@@ -367,12 +367,14 @@ where
         if index as usize >= self.num_queues {
             return Err(VhostUserError::InvalidParam);
         }
+
         // Quote from vhost-user specification:
         // Client must start ring upon receiving a kick (that is, detecting
         // that file descriptor is readable) on the descriptor specified by
         // VHOST_USER_SET_VRING_KICK, and stop ring upon receiving
         // VHOST_USER_GET_VRING_BASE.
         self.vrings[index as usize].set_queue_ready(false);
+
         if let Some(fd) = self.vrings[index as usize].get_ref().get_kick() {
             for (thread_index, queues_mask) in self.queues_per_thread.iter().enumerate() {
                 let shifted_queues_mask = queues_mask >> index;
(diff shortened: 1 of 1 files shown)


germag:
In my previous PR (#154) I was wrong about DPDK[1], at the end it calls `vring_invalidate(dev, vq);` that does something similar to `Queue::reset()`, however, I still think it is not the right place to call reset, I think it is incorrectly assumed that `GET_VRING_BASE` will always be the last message.

[1] https://github.com/DPDK/dpdk/blob/d03446724972d2a1bb645ce7f3e64f5ef0203d61/lib/vhost/vhost_user.c#L2133

stefano-garzarella (approved):
> In my previous PR (#154) I was wrong about DPDK[1], at the end it calls `vring_invalidate(dev, vq);` that does something similar to `Queue::reset()`,

Thanks for checking!

> however, I still think it is not the right place to call reset, I think it is incorrectly assumed that `GET_VRING_BASE` will always be the last message.

Yep, I agree. I think multiple calls of `GET_VRING_BASE` should return the same index.
I think we should support `VHOST_USER_RESET_DEVICE`, in the meantime maybe a well place could be `VHOST_USER_RESET_OWNER`, which itself should be a NOP, but I see [in QEMU that it is used to reset the device](https://gitlab.com/qemu-project/qemu/-/blob/master/hw/virtio/vhost-user.c#L1472) when `VHOST_USER_PROTOCOL_F_RESET_DEVICE` has not been negotiated.

stefano-garzarella:
@jiangliu @sboeuf can you take a look?
And eventually test with your code to be sure we aren't breaking anything?

jiangliu (approved):

Issue #164, opened 2023-06-01
Potential regression for vDPA support


likebreath:
With upgrading Cloud Hypervisor's dependencies of various rust-vmm crates [1], our CI reported the vDPA integration tests (that uses in-kernel vDPA device simulators) are failing with the latest release vhost v0.7.0 [2]. 

I found the breaking commit is 7e76859b3e272e64bd31cd0a63a8709c6efdfc4f, which essentially switches to use host virtual address (HVA) instead of guest physical address (GPA) with the ioctl `VHOST_SET_VRING_ADDR`. Considering the in-kernel vDPA device simulators were working fine with GPA, I wonder if this commit brings regressions to vDPA support?

[1] https://github.com/cloud-hypervisor/cloud-hypervisor/pull/5451
[2] https://cloud-hypervisor-jenkins.westus.cloudapp.azure.com/blue/organizations/jenkins/cloud-hypervisor/detail/PR-5451/6/pipeline/169

/cc @sboeuf @rbradford

sboeuf:
@stefano-garzarella do you know if the vdpa backend expects GPAs or HVAs from `VHOST_SET_VRING_ADDR`?

stefano-garzarella:
@likebreath @sboeuf vdpa backend expects IOVA that the VMM should map with `dma_map`, so maybe adding an extra mapping could fix the issue, or we should override the `set_vring_addr` .

IIUC QEMU uses HVA for vhost devices and GPA for vdpa devices, because it already maps the entire guest memory, so in this way you don't need and extra maps.

I'm not sure about the extra mapping, since the address may collide.

sboeuf:
Thanks @stefano-garzarella
IIUC, the problem is at the specification level (but there's no real specification for vhost unfortunately). It's wrong that `VHOST_SET_VRING_ADDR` expects different types of addresses depending on the backend (`vhost-net`/`vhost-vsock` vs `vdpa`).

Anyway, I agree that overriding `set_vring_addr()` implementation for `vdpa` is the best we can do here.

stefano-garzarella:
> Thanks @stefano-garzarella IIUC, the problem is at the specification level (but there's no real specification for vhost unfortunately). It's wrong that `VHOST_SET_VRING_ADDR` expects different types of addresses depending on the backend (`vhost-net`/`vhost-vsock` vs `vdpa`).

Yep, I agree, but IIRC vhost supports both IOVA and HVA, while vhost-vdpa supports only IOVA.
[...]

PR #165, opened 2023-06-02
vhost: vdpa: Provide custom `set_vring_addr()` implementation

likebreath:
### Summary of the PR

Unlike other vhost backends (e.g. vhost-net and vhost-vsock), vDPA
backends can not work with host virtual address (HVA), instead they
expect I/O virtual address (IOVA). The IOVA can be mapped 1:1 with
guest physical address (GPA) when no IOMMU is involved. This is why
the default implementation of `set_vring_addr()` from Trait
`VhostBackend` is no longer working with vDPA backends. To solve this
issue, this patch provides a custom `set_vring_addr()` implementation
for Trait `VhostKernVdpa`.

Fixes: #164

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/vhost/CHANGELOG.md
+++ b/crates/vhost/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Changed
 
 ### Fixed
+- [[#165]](https://github.com/rust-vmm/vhost/pull/165) vhost: vdpa: Provide custom set_vring_addr() implementation
 
 ### Deprecated
 
--- a/crates/vhost/src/vhost_kern/vdpa.rs
+++ b/crates/vhost/src/vhost_kern/vdpa.rs
@@ -61,6 +61,34 @@ impl<AS: GuestAddressSpace> VhostKernVdpa<AS> {
             backend_features_acked,
         }
     }
+
+    /// Set the addresses for a given vring.
+    ///
+    /// # Arguments
+    /// * `queue_index` - Index of the queue to set addresses for.
+    /// * `config_data` - Vring config data, addresses of desc_table, avail_ring
+    ///     and used_ring are in the guest address space.
+    pub fn set_vring_addr(&self, queue_index: usize, config_data: &VringConfigData) -> Result<()> {
+        if !self.is_valid(config_data) {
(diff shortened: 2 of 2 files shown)


sboeuf (dismissed):

crates/vhost/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ## [Unreleased]
 
 ### Added
+- [[#165]](https://github.com/rust-vmm/vhost/pull/165) vhost: vdpa: Provide custom set_vring_addr() implementation
```
stefano-garzarella, line 5: IMHO this is more a fix then a new feature.
I would move it to the Fixed section.

crates/vhost/src/vhost_kern/vdpa.rs
```diff
@@ -61,6 +61,34 @@ impl<AS: GuestAddressSpace> VhostKernVdpa<AS> {
...
+    /// # Arguments
+    /// * `queue_index` - Index of the queue to set addresses for.
+    /// * `config_data` - Vring config data, addresses of desc_table, avail_ring
+    ///     and used_ring are in the guest address space.
+    pub fn set_vring_addr(&self, queue_index: usize, config_data: &VringConfigData) -> Result<()> {
+        if !self.is_valid(config_data) {
+            return Err(Error::InvalidQueue);
+        }
+
+        // vDPA backends expect GPA rather than HVA, so no address
```
stefano-garzarella, line 76: I would replace GPA with IOVA (that can be mapped 1:1 with GPA, but if there is an iommu can be different).
I would also update the commit description.

stefano-garzarella (commented):
Just minor comments to avoid confusion in the future, the code LGTM.

crates/vhost/src/vhost_kern/vdpa.rs
likebreath: @stefano-garzarella PR updated. PTAL. Thanks!

sboeuf (approved):

stefano-garzarella (approved):

PR #180, opened 2023-08-08
vhost-user-backend: fetch 'used' index from guest

germag:
### Summary of the PR

commit 958cdec2b8741af77b7f90214ec9c97040bf5a8a stopped GET_VRING_BASE from resetting the vring. This was required because when the VM is stopped/resumed the index will be 0 instead of the correct value.

However, after the guest driver changes, for instance, after reboot, the 'used' index should be 0. The bug fixed in that commit had the unintended side effect of setting the 'used' index to 0 after a driver change.

    
QEMU sets the 'used' index when receiving SET_VRING_ADDR, _probably_ to  make sure that the VQ is configured. Perhaps the appropriate place is  when receiving the first kick. A better solution would be to send the used index in SET_VRING_BASE, as is done when using packed VQs.

To keep compatibility with QEMU and just in case, any implementation expects the 'used' index to be set when receiving SET_VRING_ADDR let's fetch the 'used' index from the guest when receiving that message.

### DIFF
--- a/crates/vhost-user-backend/src/handler.rs
+++ b/crates/vhost-user-backend/src/handler.rs
@@ -349,6 +349,22 @@ where
             self.vrings[index as usize]
                 .set_queue_info(desc_table, avail_ring, used_ring)
                 .map_err(|_| VhostUserError::InvalidParam)?;
+
+            // SET_VRING_BASE will only restore the 'avail' index, however, after the guest driver
+            // changes, for instance, after reboot, the 'used' index should be reset to 0.
+            //
+            // So let's fetch the used index from the vring as set by the guest here to keep
+            // compatibility with the QEMU's vhost-user library just in case, any implementation
+            // expects the 'used' index to be set when receiving a SET_VRING_ADDR message.
+            //
+            // Note: I'm not sure why QEMU's vhost-user library sets the 'user' index here,
+            // _probably_ to make sure that the VQ is already configured. A better solution would
+            // be to receive the 'used' index in SET_VRING_BASE, as is done when using packed VQs.
+            let idx = self.vrings[index as usize]
+                .queue_used_idx()
+                .map_err(|_| VhostUserError::SlaveInternalError)?;
+            self.vrings[index as usize].set_queue_next_used(idx);
+
             Ok(())
         } else {
             Err(VhostUserError::InvalidParam)
--- a/crates/vhost-user-backend/src/vring.rs
+++ b/crates/vhost-user-backend/src/vring.rs
@@ -10,6 +10,7 @@ use std::io;
 use std::ops::{Deref, DerefMut};
 use std::os::unix::io::{FromRawFd, IntoRawFd};
 use std::result::Result;
+use std::sync::atomic::Ordering;
 use std::sync::{Arc, Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuard};
 
 use virtio_queue::{Error as VirtQueError, Queue, QueueT};
@@ -73,6 +74,12 @@ pub trait VringT<M: GuestAddressSpace>:
 
     /// Set queue next avail head.
     fn set_queue_next_avail(&self, base: u16);
+
+    /// Set queue next used head.
+    fn set_queue_next_used(&self, idx: u16);
+
+    /// Get queue next used head index from the guest memory.
+    fn queue_used_idx(&self) -> Result<u16, VirtQueError>;
 
     /// Set configured queue size.
     fn set_queue_size(&self, num: u16);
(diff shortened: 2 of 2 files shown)


crates/vhost-user-backend/src/vring.rs
```diff
@@ -192,6 +196,16 @@ impl<M: GuestAddressSpace> VringState<M> {
         self.queue.set_next_avail(base);
     }
 
+    /// Set the queue next used head index to the value from the used ring.
+    fn fetch_queue_next_used_from_guest(&mut self) -> Result<(), VirtQueError> {
```
sboeuf, line 200: I'd prefer a method called `set_queue_next_used()` where you would provide the expected value.
```rust
fn set_queue_next_used(&mut self, idx: u16) {
    self.queue.set_next_used(idx);
}
```
And I'd expect another one for getting the used_idx.

germag: Just a clarification, the new function to get the used idx in vring, which index should return the used idx from `self.queue` or from the guest?, I mean should it call:
1- `self.queue.next_used()`
or 
2- `self.queue.used_idx(...)`

sboeuf: Well based on your current code, I'd simply define the following:
```rust
/// Get queue used index.
fn queue_used_idx(&self) -> u16 {
    self.queue.used_idx(...)
}
```

crates/vhost-user-backend/src/vring.rs
```diff
@@ -192,6 +196,16 @@ impl<M: GuestAddressSpace> VringState<M> {
...
+        let idx = self
+            .queue
+            .used_idx(self.mem.memory().deref(), Ordering::Acquire)?;
```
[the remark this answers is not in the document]
stefano-garzarella: The spec also says: "idx field indicates where the device would put the next descriptor entry in the ring (modulo the queue size). This starts at 0, and increases."

Anyway the Linux driver allocating all memory with __GFP_ZERO resets used and avail to 0 during initialization.
However in case of an error then it resets both avail and used.

So reading it from the guest should be okay.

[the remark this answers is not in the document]
stefano-garzarella: Just checked and `vhost` kernel code does exactly the same, it reads used from the guest memory, but of course in different places:
- vhost-net when VHOST_NET_SET_BACKEND ioctl is received
- vhost-vsock when VHOST_VSOCK_SET_RUNNING ioctl is received

So I think your approach should be fine ;-)

germag:
v2:
- Replace `fetch_queue_next_used_from_guest()` by `set_queue_next_used()` and `queue_used_idx()`

crates/vhost-user-backend/src/vring.rs
germag: Done

crates/vhost-user-backend/src/vring.rs
```diff
@@ -433,6 +460,14 @@ impl<M: 'static + GuestAddressSpace> VringT<M> for VringRwLock<M> {
         self.write_lock().set_queue_next_avail(base)
     }
 
+    fn set_queue_next_used(&self, idx: u16) {
+        self.write_lock().set_queue_next_used(idx)
+    }
+
+    fn queue_used_idx(&self) -> Result<u16, VirtQueError> {
+        self.write_lock().queue_used_idx()
```
stefano-garzarella, line 468: Is `write_lock` needed here?

crates/vhost-user-backend/src/handler.rs
```diff
@@ -349,6 +349,14 @@ where
             self.vrings[index as usize]
                 .set_queue_info(desc_table, avail_ring, used_ring)
                 .map_err(|_| VhostUserError::InvalidParam)?;
+
+            // SET_VRING_BASE will only restore the avail index, so we need to fetch the used index
+            // from the vring as set by the guest
```
stefano-garzarella, line 354: I would suggest adding some more info included in the commit description here as well, such as the behavior of QEMU

crates/vhost-user-backend/src/vring.rs
germag: Opss true, dumb copy-paste error, thanks!

crates/vhost-user-backend/src/handler.rs
germag: Ok

germag:
v3:
- remove `.write_lock()`
- improve code comment

stefano-garzarella (approved):

sboeuf (approved):

stefano-garzarella:
I think we need to release this fix ASAP with v0.10.1 since I had the same issue also with vhost-device-vsock.

I'm going to prepare the release.

@sboeuf @jiangliu @slp @eryugey WDYT?

stefano-garzarella:
FYI PR opened here with the release: https://github.com/rust-vmm/vhost/pull/183

We can discuss there.

PR #183, opened 2023-08-21
vhost-user-backend: release v0.10.1

stefano-garzarella:
Commit 8a4ba9d ("vhost-user-backend: fetch 'used' index from guest") fixes an issue introduced in v0.10.0 that affects all vhost-user backends like virtiofsd [1] or rust-vmm's vhost-device. I easily reproduced the problem with vhost-device-vsock as well: just restart the guest kernel and the device no longer works.

[1] https://gitlab.com/virtio-fs/virtiofsd/-/issues/120

### DIFF
--- a/crates/vhost-user-backend/CHANGELOG.md
+++ b/crates/vhost-user-backend/CHANGELOG.md
@@ -9,6 +9,11 @@
 
 ### Deprecated
 
+## v0.10.1
+
+### Fixed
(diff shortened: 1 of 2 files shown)


sboeuf (approved):

slp (approved):
LGTM

jiangliu (approved):

PR #187, opened 2023-09-08
Clean master slave

aesteve-rh:
### Summary of the PR

Change uses of master/slave for backend/frontend in the codebase.

### Requirements

requirements:

  description line.
  test.


sboeuf (approved):
LGTM

crates/vhost/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 
 ## [0.8.1]
 
+### Changed
```
stefano-garzarella, line 14: We already release 0.8.1, these changes should go under the `## [Unreleased]` section

crates/vhost-user-backend/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 
 ## v0.10.1
 
+### Changed
+- Change uses of master/slave for frontend/backend in the codebase.
```
stefano-garzarella, line 15: The same here.

crates/vhost/src/vhost_user/message.rs
```diff
@@ -941,30 +941,30 @@ bitflags! {
     }
 }
 
-/// Max entries in one virtio-fs slave request.
+/// Max entries in one virtio-fs backend request.
 pub const VHOST_USER_FS_SLAVE_ENTRIES: usize = 8;
```
stefano-garzarella, line 945: What about updating also the name of this constant?
```suggestion
pub const VHOST_USER_FS_BACKEND_ENTRIES: usize = 8;
```

stefano-garzarella:
I left some minor comments, but it LGTM ;-)

Since it is a breaking change, @jiangliu @eryugey @slp do you have any thought on this?

aesteve-rh:
> Since it is a breaking change

Yeah... I can help to update other repositories consuming this API that would be broken after the update.
What would be the best way/order to proceed for those?

crates/vhost/src/vhost_user/message.rs
aesteve-rh: Ah! Forgot about the all-uppercase case, I'll change this and check for other occurrences.

stefano-garzarella:
> Yeah... I can help to update other repositories consuming this API that would be broken after the update.

Thanks! Really appreciated!

> What would be the best way/order to proceed for those?

I think we need to do publish a new release of `vhost` and `vhost-user-backed` as a first step. Since both are still `0.y.z`, a new `y` release should be fine since it is a major release. Then we can start to update dependencies like `vhost-device`'s crates, `virtiofsd`, etc.

stefano-garzarella (approved):

sboeuf (approved):

rbradford (approved):

aesteve-rh:
@stefano-garzarella thanks!

I'll keep an eye on the next release and start updating other repositories once available (probably will start with `vhost-device`, as I already have a fork).

crates/vhost-user-backend/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ### Added
 
 ### Changed
+- Change uses of master/slave for frontend/backend in the codebase.
```
Ablu, line 7: Maybe we should also point out that the features got renamed? That just tripped me off because `cargo update` silently refuses to update. Only when forcing the version I learn about the reason:

```
❯ cargo update -p vhost@0.8.0 --precise 0.8.1
    Updating crates.io index
error: failed to select a version for `vhost`.
    ... required by package `vhost-device-gpio v0.1.0 (/home/ablu/projects/rust-vmm/vhost-device/crates/vhost-device-gpio)`
versions that meet the requirements `^0.8` are: 0.8.1

the package `vhost-device-gpio` depends on `vhost`, with features: `vhost-user-slave` but `vhost` does not have these features.


failed to select a version for `vhost` which could resolve this conflict
```

stefano-garzarella: Yeah, good point. We should also add a link to this PR as we usually do in the changelog

stefano-garzarella: I'm going to open a Draft PR for the next release, and I'll include this change there, so we don't forget it!

stefano-garzarella: @aesteve-rh don't worry, we need to update the CHANGELOG in any case, so I can handle that in that PR. I'll post soon a draft and ask you and @Ablu to review ;-)

stefano-garzarella: @Ablu just to be sure, we need to put that change in the `vhost` changelog, not this one (`vhost-user-backend`), right?

Ablu: > @Ablu just to be sure, we need to put that change in the `vhost` changelog, not this one (`vhost-user-backend`), right?

Right. I find it a bit confusing that there is a backend feature for `vhost` but also a `vhost-user-backend` crate... But I have not digged into what exactly happens within `vhost` there.

stefano-garzarella: @Ablu eheh this is another good point, but I don't know how we can fix without a disruptive change :-(

Ablu: > @Ablu eheh this is another good point, but I don't know how we can fix without a disruptive change :-(

Yeah, for now just documenting it in `vhost`'s CHANGELOG should be fine. The other thing is unrelated anyway.

stefano-garzarella: Handling this in https://github.com/rust-vmm/vhost/pull/193

PR #192, opened 2023-09-28
vhost-user-backend: remove return value from handle_event

### LINKED ISSUE #144: Clarification of return type of `handle_event`
`VhostUserBackendMut::handle_event` has a bool return type that is not documented. Drilling into the code reveals that 
```
                // handle_event() returns true if an event is received from the exit event fd.
```
[1]

However, it looks like this is handled at [2]. Are there still valid use-cases where someone else should return a true there? Or shall we drop the bool from `VhostUserBackend` and `VhostUserBackendMut`?

@jiangliu: Maybe you can comment on this?

[1] https://github.com/rust-vmm/vhost/blame/40006d0b3981504320ae66e165c9fd36ce9cfb18/crates/vhost-user-backend/src/event_loop.rs#L196
[2] https://github.com/rust-vmm/vhost/blame/40006d0b3981504320ae66e165c9fd36ce9cfb18/crates/vhost-user-backend/src/event_loop.rs#L207

stefano-garzarella:
### Summary of the PR

The return value of VhostUserBackend::handle_event() is undocumented and difficult to interpret.
The current implementation used it to interrupt the event loop as it does when we receive an exit event.

All current implementations checked (rust-vmm/vhost-device, virtiofsd) return an error or always false, effectively not using this feature.

Since we already have a mechanism for breaking the event loop, we can avoid this ambiguous and redundant feature.

Closes #144

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/vhost-user-backend/CHANGELOG.md
+++ b/crates/vhost-user-backend/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Changed
 - Change uses of master/slave for frontend/backend in the codebase.
+- [[#192]](https://github.com/rust-vmm/vhost/pull/192) vhost-user-backend: remove return value from handle_event
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


sboeuf (approved):

slp (approved):
LGTM.

PR #196, opened 2023-10-07
Use Cow<'static, str> for VhostUserDaemon name field.

epilys:
### Summary of the PR

Change the name's type `String` to `Cow<'static, str>`. This allows the
library user to prevent string allocation from static strings, which is
how most crates use it.

This is an API breaking change.

### Requirements

requirements:

  description line.
  test.


crates/vhost-user-backend/src/lib.rs
```diff
@@ -124,7 +125,7 @@ where
         mut handler: BackendReqHandler<Mutex<VhostUserHandler<T>>>,
     ) -> Result<()> {
         let handle = thread::Builder::new()
-            .name(self.name.clone())
+            .name(self.name.to_string())
```
Ablu, line 128: Hm... This is the only use and requires ownership. So should we maybe just drop it from the daemon struct and make it part of the start_daemon function (as a String that can be moved through)?

Though, overall, I think this is a lot of code churn (for everyone using this) and additional complexity for very little gain.

PR #202, opened 2023-11-14
Proper enum <-> value conversion

germag:
### Summary of the PR
`NOOP` and `MAX_CMD` are not present in the vhost-user protocol, these are use only to verify that the value is correct before transmuting it to an `enum` variant.
    
This has two problems, firstly it exposes of non-standard values in a public interface, this internal implementation detail could be confusing, and secondly it assumes that the values are consecutive with no "holes" between them.
    
This results in having to define requests that are not actually supported, and error-prone, which can generate UB when transmuting.
    
Instead, it is better to implement `TryFrom<u32>` for enums that need to convert between variants and u32. This allows to remove the unsafe transmute.

### DIFF
--- a/crates/vhost/src/vhost_user/message.rs
+++ b/crates/vhost/src/vhost_user/message.rs
@@ -55,150 +55,167 @@ pub(super) trait Req:
     fn is_valid(value: u32) -> bool;
 }
 
-/// Type of requests sending from frontends to backends.
-#[repr(u32)]
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub enum FrontendReq {
-    /// Null operation.
-    NOOP = 0,
-    /// Get from the underlying vhost implementation the features bit mask.
-    GET_FEATURES = 1,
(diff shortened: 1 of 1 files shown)


crates/vhost/src/vhost_user/message.rs
```diff
@@ -55,150 +55,167 @@ pub(super) trait Req:
...
+        }
+    ) => {
+        #[repr($T)]
+        $(#[$meta])*
+        $vis enum $enum {
+            $($(#[$variant_meta])* $variant $(= $val)?,)*
+        }
+
+        impl std::convert::TryFrom<$T> for $enum {
+            type Error = ();
```
stefano-garzarella, line 74: Should we make it parametric or use something like std::num::TryFromIntError ?

stefano-garzarella (approved):
Very nice cleanup! Thanks!

I just left a very minor comment, LGTM!

crates/vhost/src/vhost_user/message.rs
stefano-garzarella: It looks like there is no way to instantiate a std::num::TryFromIntError

PR #203, opened 2023-11-14
Add back-end's internal state migration support

germag:
### Summary of the PR
  
A new set of messages have been introduced in the vhost-user protocol to support the migration of devices with back-end's internal state[0].
    
These are the additions to the protocol:
    - New vhost-user protocol feature `VHOST_USER_PROTOCOL_F_DEVICE_STATE`
    - `SET_DEVICE_STATE_FD` function: Front-end and back-end negotiate a file
      descriptor over which to transfer the state. It also includes
      establishing the direction of transfer and migration phase.
    - `CHECK_DEVICE_STATE`: After the state has been transferred through the
      file descriptor, the front-end invokes this function to verify
      success.  There is no in-band way (through the file descriptor) to
      indicate failure, so we need to check explicitly.
    
[0] https://github.com/qemu/qemu/commit/019233096c03b826e0e677115b6e3c550a54a48d

This code is part of the effort to support live migration in virtiofsd and is currently being used in a version of virtiofsd that supports migration: https://gitlab.com/virtiofsd-live-migration

### Note to reviewers

~Please do not merge this PR yet, this PR depends on https://github.com/rust-vmm/vhost/pull/202, because by adding `FrontendReq::SET_DEVICE_STATE_FD` a "hole" was created in value `41`, this makes the APi unsound, since `VhostUserMsgHeader::get_code()` call `transmute_copy()` which assumes that the values are continuous and have no "holes" between them.~

~I'll rebase and adjust the code after PR 202 gets merged.~

### DIFF
--- a/vhost/src/vhost_user/message.rs
+++ b/vhost/src/vhost_user/message.rs
@@ -440,6 +440,8 @@ bitflags! {
         const STATUS = 0x0001_0000;
         /// Support Xen mmap.
         const XEN_MMAP = 0x0002_0000;
+        /// Support transferring internal device state.
+        const DEVICE_STATE = 0x0008_0000;
     }
 }
 
--- a/vhost/src/vhost_user/message.rs
+++ b/vhost/src/vhost_user/message.rs
@@ -178,6 +178,12 @@ enum_value! {
         /// Query the backend for its device status as defined in the VIRTIO
         /// specification.
         GET_STATUS = 40,
+        /// Begin transfer of internal state to/from the backend for migration
+        /// purposes.
+        SET_DEVICE_STATE_FD = 42,
+        /// After transferring state, check the backend for any errors that may have
+        /// occurred during the transfer
+        CHECK_DEVICE_STATE = 43,
     }
 }
 
@@ -907,6 +913,61 @@ impl VhostUserMsgValidator for VhostUserLog {
     }
 }
 
+enum_value! {
+    /// Direction of state transfer for migration
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+    pub enum VhostTransferStateDirection: u32 {
+        /// Outgoing migration: Transfer state from back-end to front-end
+        SAVE = 0,
+        /// Incoming migration: Transfer state from front-end to back-end
+        LOAD = 1,
+    }
+}
+
+enum_value! {
+    /// Migration phases during which state transfer can occur
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+    pub enum VhostTransferStatePhase: u32 {
+        /// The device (and all its vrings) are stopped
+        STOPPED = 0,
+    }
+}
+
+/// Query/send virtio-fs migration state
+// Note: this struct is not defined as `packed` in the SPEC and although
+// it is not necessary, since the struct has no padding, it simplifies
+// reviewing it because it is a requirement for implementing `ByteValued`.
+#[repr(C, packed)]
+#[derive(Clone, Copy, Default)]
+pub struct VhostUserTransferDeviceState {
+    /// Direction of state transfer (save/load)
+    pub direction: u32,
+    /// Migration phase during which the transfer takes place
+    pub phase: u32,
+}
+
+// SAFETY: Safe because VhostUserTransferDeviceState is a POD
+// (i.e., none of its fields are references or raw pointers),
+// and there is no compiler-inserted padding.
+unsafe impl ByteValued for VhostUserTransferDeviceState {}
+
+impl VhostUserTransferDeviceState {
+    /// Create a new instance.
+    pub fn new(direction: VhostTransferStateDirection, phase: VhostTransferStatePhase) -> Self {
+        VhostUserTransferDeviceState {
+            direction: direction as u32,
+            phase: phase as u32,
+        }
+    }
+}
+
+impl VhostUserMsgValidator for VhostUserTransferDeviceState {
+    fn is_valid(&self) -> bool {
+        VhostTransferStateDirection::try_from(self.direction).is_ok()
+            && VhostTransferStatePhase::try_from(self.phase).is_ok()
+    }
+}
+
 // Bit mask for flags in virtio-fs backend messages
 bitflags! {
     #[derive(Copy, Clone, Debug, Eq, PartialEq, Default)]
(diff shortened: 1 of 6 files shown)


crates/vhost-user-backend/src/backend.rs
```diff
@@ -110,6 +113,35 @@ pub trait VhostUserBackend: Send + Sync {
         vrings: &[Self::Vring],
         thread_id: usize,
     ) -> Result<()>;
+
+    /// Initiate transfer of internal state for the purpose of migration to/from the back-end.
+    ///
+    /// Depending on `direction`, the state should either be saved (i.e. serialized and written to
+    /// `file`) or loaded (i.e. read from `file` and deserialized).  Note that this function must
+    /// not block during transfer, i.e. I/O to/from `file` must be done outside of this function.
```
XanClic, line 121: Maybe we should add documentation on the return type; without reading the vhost-user doc, it isn’t obvious why a `File` can be returned. Something like “The back-end can choose to use a different channel than `file`. If so, it must return a `File` that the front-end can use.”?

crates/vhost-user-backend/src/backend.rs
```diff
@@ -384,6 +473,21 @@ impl<T: VhostUserBackendMut> VhostUserBackend for RwLock<T> {
...
+        phase: VhostTransferStatePhase,
+        file: File,
+    ) -> Result<Option<File>> {
+        self.write()
+            .unwrap()
+            .set_device_state_fd(direction, phase, file)
+    }
+
+    fn check_device_state(&self) -> Result<()> {
+        self.write().unwrap().check_device_state()
```
XanClic, line 489: Again my fault (like the missing comment), but won’t `read()` suffice here? `VhostUserBackendMut::check_device_state()` only takes an immutable `&self`.

crates/vhost-user-backend/src/backend.rs
```diff
@@ -110,6 +113,35 @@ pub trait VhostUserBackend: Send + Sync {
...
+            "back end does not support state transfer",
+        ))
+    }
+
+    /// After transferring internal state, check for any resulting errors, including potential
+    /// deserialization errors when loading state.
+    ///
+    /// Although this function return a `Result`, the front-end will not receive any details about
+    /// this error.
+    fn check_device_state(&self) -> Result<()> {
```
XanClic, line 139: Personally, I liked using `std::result::Result<(), ()>` to make it explicit in typing that the error message would be lost, because the function/message name sounds like the front-end might want to inquire about the exact nature of the error (and documentation is easy to miss).
But then again, most messages allow you to return a non-trivial error, and I don’t think any of them will return anything more than “error/success” to the front-end, so really, there is no point in treating this any differently from those other messages.

crates/vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -519,6 +548,44 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
...
+                    }
+                    Err(_) => {
+                        let msg = VhostUserU64::new(0x101);
+                        self.main_sock.send_message(&reply_hdr, &msg, None)?;
+                    }
+                }
+            }
+            Ok(FrontendReq::CHECK_DEVICE_STATE) => {
+                let res = self.backend.check_device_state();
+                self.send_ack_message(&hdr, res)?;
```
XanClic, line 587: As far as I can see, `send_ack_message()` only sends a reply if `F_REPLY_ACK` has been negotiated. However, `CHECK_DEVICE_STATE` is defined to always send a reply, independently of `F_REPLY_ACK`. So I don’t think we should use `send_ack_message()` here (or if we do, I think it would need an additional parameter to force it to send the reply message).

XanClic:
The compiled version of the documentation linked in patch 1 is here: https://qemu-project.gitlab.io/qemu/interop/vhost-user.html

crates/vhost-user-backend/src/backend.rs
germag: Yes, I plan to add more docs (for that reason I leave the [Draft])

crates/vhost-user-backend/src/backend.rs
germag: Yes, I missed that one

crates/vhost/src/vhost_user/backend_req_handler.rs
germag: True

crates/vhost-user-backend/src/backend.rs
germag: Nop, it's a clippy warn:
https://rust-lang.github.io/rust-clippy/master/index.html#/result_unit_err

crates/vhost-user-backend/src/backend.rs
XanClic: (Looks good to me in the current version. (github won’t let me resolve the conversation, because only the PR author and anyone with write access to the repo can do so, hence this comment.))

crates/vhost-user-backend/src/backend.rs
XanClic: Interesting, for some reason I couldn’t reproduce that warning (and didn’t find anything about it). Ah well.

crates/vhost/src/vhost_user/backend_req_handler.rs
XanClic: (Looks good in the current version.)

XanClic:
Looks good to me, thanks!

stefano-garzarella (approved):
nit: I see we are using `back end`, `back-end`, and `backend` in this PR, but also in the code that is already merged, so I think we should decide which term to use (maybe `back-end` to follow the spec) and use it all over the place (not for this PR).

vhost/src/vhost_user/message.rs
```diff
@@ -907,6 +915,56 @@ impl VhostUserMsgValidator for VhostUserLog {
...
+    /// Migration phases during which state transfer can occur
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+    pub enum VhostTransferStatePhase: u32 {
+        /// The device (and all its vrings) are stopped
+        STOPPED = 0,
+    }
+}
+
+/// Query/send virtio-fs migration state
+#[repr(C)]
```
Ablu, line 939: I would add a packed here to simplify reviews of unsafe `ByteValued` impls.

vhost/src/vhost_user/message.rs
```diff
@@ -907,6 +913,56 @@ impl VhostUserMsgValidator for VhostUserLog {
...
+#[derive(Clone, Copy, Default)]
+pub struct VhostUserTransferDeviceState {
+    /// Direction of state transfer (save/load)
+    pub direction: u32,
+    /// Migration phase during which the transfer takes place
+    pub phase: u32,
+}
+
+// SAFETY: Safe because all fields of VhostUserTransferDeviceState are POD.
```
Ablu, line 948: POD is only one requirement. Since we seem to got this wrong in a few places we should also mention the packed requirement here IMHO. What do you think?

Ablu (commented):
LGTM. Only some nit-picking on the ByteValued. I think since it has been source of confusion before, we should probably be a bit more explicit about it's requirements.

germag:
v2:
- Add comments in message definition

vhost/src/vhost_user/message.rs
germag: Done

germag:
Rebase

PR #210, opened 2023-11-17
vhost-user-backend: SET_FEATURES must not disable any ring

germag:
### Summary of the PR

The vhost-user spec has been updated to clarify enabling and disabling vrings. The current spec states that upon receiving a `VHOST_USER_SET_FEATURES` message from the front-end without `VHOST_USER_F_PROTOCOL_FEATURES` set, the back-end must enable all rings immediately. Also, while processing the rings (whether they are enabled or not), the back-end must support changing some configuration aspects on the fly.

However, the vhost-user-backend implementation will disable all vring whenever this feature is negotiated, which means that every `VHOST_USER_SET_FEATURES` message could unintentionally halt the device.

This is an incorrect interpretation of the spec, and is problematic because the `VHOST_F_LOG_ALL` feature is also set or cleared by `VHOST_USER_SET_FEATURES`, which happens during migration. And doing so should not disable any vring nor halt the device.

This code is part of the effort to support live migration in virtiofsd and is currently being used in a version of virtiofsd that supports migration: https://gitlab.com/virtiofsd-live-migration

### DIFF
--- a/crates/vhost-user-backend/src/handler.rs
+++ b/crates/vhost-user-backend/src/handler.rs
@@ -200,6 +200,7 @@ where
             }
         }
 
+        self.vrings[index as usize].set_enabled(false);
         self.vrings[index as usize].set_queue_ready(true);
 
         Ok(())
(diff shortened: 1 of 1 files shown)


Ablu (approved):
Tests would be great, but there seem to be little tests for specifically this in general... So adding them just for this is maybe a bit overkill.

stefano-garzarella (approved):
LGTM! Thanks for fixing this issue!

PR #212, opened 2023-11-20
vhost-user-backend: check `index` in set_vring_base()

### LINKED ISSUE #211: Missing Input Parameter Validation
From @y-x41

While reviewing the vhost crate, it was found that the function VHostUserHandler::set_vring_base() lacks validation of the input parameter index, which is used for indexing an internal vector and could potentially lead to a Out-of-Bounds write resulting in an application panic. The below listing depicts the vulnerable code segment.

```rust
impl<S, V, B> VhostUserBackendReqHandlerMut for VhostUserHandler<S, V, B>
where
    S: VhostUserBackend<V, B>,
    V: VringT<GM<B>>,
    B: NewBitmap + Clone,
{

    // [...]

    fn set_vring_base(&mut self, index: u32, base: u32) -> VhostUserResult<()> {
        let event_idx: bool = (self.acked_features & (1 << VIRTIO_RING_F_EVENT_IDX)) != 0;

        self.vrings[index as usize].set_queue_next_avail(base as u16);
        self.vrings[index as usize].set_queue_event_idx(event_idx);
        self.backend.set_event_idx(event_idx);

        Ok(())
    }
}
```

X41 advises performing thorough input parameter validation to prevent any possibility of encountering potential Out-of-Bounds read/writes.

stefano-garzarella:
VHostUserHandler::set_vring_base() lacks validation of the input parameter `index`, which is used for indexing an internal vector and could potentially lead to a Out-of-Bounds write resulting in a panic.

Closes #211

### DIFF
--- a/vhost-user-backend/src/handler.rs
+++ b/vhost-user-backend/src/handler.rs
@@ -369,6 +369,10 @@ where
     }
 
     fn set_vring_base(&mut self, index: u32, base: u32) -> VhostUserResult<()> {
+        if index as usize >= self.num_queues {
+            return Err(VhostUserError::InvalidParam);
+        }
+
         let event_idx: bool = (self.acked_features & (1 << VIRTIO_RING_F_EVENT_IDX)) != 0;
 
         self.vrings[index as usize].set_queue_next_avail(base as u16);


crates/vhost-user-backend/src/handler.rs
```diff
@@ -367,6 +367,10 @@ where
     }
 
     fn set_vring_base(&mut self, index: u32, base: u32) -> VhostUserResult<()> {
+        if index as usize >= self.num_queues {
```
Ablu, line 370: My first attempt would have been: `let [mut] vring = self.vrings.get(index).ok_or_else(...)?`?

Removes the double indexing and does not rely on some variable being in sync with the vector.

stefano-garzarella: I honestly did direct copy and paste from the other methods that all do this.

That said, I like your approach much better, but I would send another patch in another PR where we change them all together. What do you think?

stefano-garzarella: I did it here: https://github.com/rust-vmm/vhost/pull/214

I don't know whether to merge this patch with that at this point and just have the other PR...

Ablu: I am confused, the other PR now lists this as dependency, but this is not updated? :thinking: Got no opinion on doing this through one or two PRs :)

stefano-garzarella: @Ablu sorry ;-)
I mean, the other PR is based on this (it includes the patch in this PR). Now my doubt is whether to have 2 patches as now (the first one is then overwritten by the second one, but it could be used for some backport), or to have directly only the second patch (the one added in the other PR).

Ablu: Ah! Either sounds good to me.

Ablu (approved):

crates/vhost-user-backend/src/handler.rs
stefano-garzarella: Okay, let's leave them separate for now ;-)

Ablu:
@stefano-garzarella: Github claims conflicts that require manual resolving here.

stefano-garzarella:
> @stefano-garzarella: Github claims conflicts that require manual resolving here.

Yeah, just rebase.

jiangliu (approved):

Ablu (approved):

PR #214, opened 2023-11-20
vhost-user-backend: take reference to `vring` once

stefano-garzarella:
vhost-user-backend: take reference to `vring` once
    
In many places we access the `VhostUserHandler::vrings` array many
times and at the beginning of each function we check the index to
avoid OOB.
    
It happened that we forgot about this check (see #211), so let's
avoid this error-prone pattern by taking the reference to the vring
using `Vec::get()`, and avoiding multiple indexing.
    
Suggested-by: Erik Schilling <erik.schilling@linaro.org>
Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>

### DIFF
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 76.3,
+  "coverage_score": 75.0,
   "exclude_path": "vhost/src/vhost_kern/",
   "crate_features": "vhost/vhost-user-frontend,vhost/vhost-user-backend"
 }
--- a/vhost-user-backend/src/handler.rs
+++ b/vhost-user-backend/src/handler.rs
@@ -200,8 +200,8 @@ where
             }
         }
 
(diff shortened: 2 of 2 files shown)


stefano-garzarella:
Depends on #212

Ablu (dismissed):

Ablu:
Thanks :)

stefano-garzarella:
> Thanks :)

Thank you for the inspiration ;-)

I don't know why, but this change reduces the coverage, mmm. I'll update the score.

Ablu:
> I don't know why, but this change reduces the coverage, mmm. I'll update the score.

Probably the error case never got exercised. But I would argue that it was an issue with the existing tests and the previous coverage was just "wrong" :). So adjusting them sounds fine to me!

stefano-garzarella:
#212 is now merged, so I rebased on main and marged ready

Ablu (approved):

PR #218, opened 2023-12-15
Adding POSTCOPY support

ShadowCurse:
### Summary of the PR

Added ability to use `VHOST_USER_POSTCOPY_ADVISE`, `VHOST_USER_POSTCOPY_LISTEN` and `VHOST_USER_POSTCOPY_END` messages. This includes sending them from the frontend and processing them on the backend.

The `POSTCOPY` messages are only usable when
`VHOST_USER_PROTOCOL_F_PAGEFAULT` feature is negotiated.

Messages and their descriptions are:
- `VHOST_USER_POSTCOPY_ADVISE`:
    When the front-end sends this message to the backend,
    the back-end must open a userfaultfd for later use
    and send it's fd to the front-end.

- `VHOST_USER_POSTCOPY_LISTEN`
    When the back-end receives this message it must ensure
    that shared memory is registered with userfaultfd to
    cause faulting of non-present pages.
    This is always sent sometime after a `VHOST_USER_POSTCOPY_ADVISE`.

- `VHOST_USER_POSTCOPY_END`
    When the back-end receives this message it must disable the
    userfaultfd. The reply is an acknowledgement only.

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.

### DIFF
--- a/.buildkite/rust-vmm-ci-tests.json
+++ b/.buildkite/rust-vmm-ci-tests.json
@@ -30,7 +30,7 @@
     },
     {
       "test_name": "unittests-gnu-all-without-xen",
-      "command": "cargo test --workspace --no-default-features --features test-utils,vhost-vsock,vhost-kern,vhost-vdpa,vhost-net,vhost-user,vhost-user-frontend,vhost-user-backend",
+      "command": "cargo test --workspace --no-default-features --features test-utils,vhost-vsock,vhost-kern,vhost-vdpa,vhost-net,vhost-user,vhost-user-frontend,vhost-user-backend,postcopy",
       "platform": [
         "x86_64",
         "aarch64"
@@ -46,7 +46,7 @@
     },
     {
       "test_name": "unittests-musl-all-without-xen",
-      "command": "cargo test --workspace --target {target_platform}-unknown-linux-musl --no-default-features --features test-utils,vhost-vsock,vhost-kern,vhost-vdpa,vhost-net,vhost-user,vhost-user-frontend,vhost-user-backend",
+      "command": "cargo test --workspace --target {target_platform}-unknown-linux-musl --no-default-features --features test-utils,vhost-vsock,vhost-kern,vhost-vdpa,vhost-net,vhost-user,vhost-user-frontend,vhost-user-backend,postcopy",
       "platform": [
         "x86_64",
         "aarch64"
@@ -62,7 +62,7 @@
     },
     {
       "test_name": "clippy-all-without-xen",
-      "command": "cargo clippy --workspace --bins --examples --benches --all-targets --no-default-features --features test-utils,vhost-vsock,vhost-kern,vhost-vdpa,vhost-net,vhost-user,vhost-user-frontend,vhost-user-backend -- -D warnings -D clippy::undocumented_unsafe_blocks",
+      "command": "cargo clippy --workspace --bins --examples --benches --all-targets --no-default-features --features test-utils,vhost-vsock,vhost-kern,vhost-vdpa,vhost-net,vhost-user,vhost-user-frontend,vhost-user-backend,postcopy -- -D warnings -D clippy::undocumented_unsafe_blocks",
       "platform": [
         "x86_64",
         "aarch64"
@@ -78,7 +78,7 @@
     },
     {
       "test_name": "check-warnings-all-without-xen",
-      "command": "RUSTFLAGS=\"-D warnings\" cargo check --all-targets --workspace --no-default-features --features test-utils,vhost-vsock,vhost-kern,vhost-vdpa,vhost-net,vhost-user,vhost-user-frontend,vhost-user-backend",
+      "command": "RUSTFLAGS=\"-D warnings\" cargo check --all-targets --workspace --no-default-features --features test-utils,vhost-vsock,vhost-kern,vhost-vdpa,vhost-net,vhost-user,vhost-user-frontend,vhost-user-backend,postcopy",
       "platform": [
         "x86_64",
         "aarch64"
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
   "coverage_score": 72.12,
   "exclude_path": "vhost/src/vhost_kern/",
-  "crate_features": "vhost/vhost-user-frontend,vhost/vhost-user-backend"
+  "crate_features": "vhost/vhost-user-frontend,vhost/vhost-user-backend,vhost-user-backend/postcopy"
 }
--- a/vhost-user-backend/Cargo.toml
+++ b/vhost-user-backend/Cargo.toml
@@ -10,10 +10,12 @@ license = "Apache-2.0"
 
 [features]
 xen = ["vm-memory/xen", "vhost/xen"]
+postcopy = ["vhost/postcopy", "userfaultfd"]
 
 [dependencies]
 libc = "0.2.39"
 log = "0.4.17"
+userfaultfd = { version = "0.7.0", optional = true }
 vhost = { path = "../vhost", version = "0.10", features = ["vhost-user-backend"] }
 virtio-bindings = "0.2.1"
 virtio-queue = "0.11.0"
--- a/vhost-user-backend/src/handler.rs
+++ b/vhost-user-backend/src/handler.rs
@@ -6,10 +6,14 @@
 use std::error;
 use std::fs::File;
 use std::io;
+#[cfg(feature = "postcopy")]
+use std::os::fd::FromRawFd;
 use std::os::unix::io::AsRawFd;
 use std::sync::Arc;
 use std::thread;
 
+#[cfg(feature = "postcopy")]
+use userfaultfd::{Uffd, UffdBuilder};
 use vhost::vhost_user::message::{
     VhostTransferStateDirection, VhostTransferStatePhase, VhostUserConfigFlags,
     VhostUserMemoryRegion, VhostUserProtocolFeatures, VhostUserSingleMemoryRegion,
@@ -71,6 +75,8 @@ impl error::Error for VhostUserHandlerError {}
 pub type VhostUserHandlerResult<T> = std::result::Result<T, VhostUserHandlerError>;
 
 struct AddrMapping {
+    #[cfg(feature = "postcopy")]
+    local_addr: u64,
     vmm_addr: u64,
     size: u64,
     gpa_base: u64,
@@ -89,6 +95,8 @@ pub struct VhostUserHandler<T: VhostUserBackend> {
     mappings: Vec<AddrMapping>,
     atomic_mem: GM<T::Bitmap>,
     vrings: Vec<T::Vring>,
+    #[cfg(feature = "postcopy")]
+    uffd: Option<Uffd>,
     worker_threads: Vec<thread::JoinHandle<VringEpollResult<()>>>,
 }
 
@@ -148,6 +156,8 @@ where
             mappings: Vec::new(),
             atomic_mem,
             vrings,
+            #[cfg(feature = "postcopy")]
+            uffd: None,
             worker_threads,
         })
     }
@@ -280,20 +290,21 @@ where
         let mut mappings: Vec<AddrMapping> = Vec::new();
 
         for (region, file) in ctx.iter().zip(files) {
-            regions.push(
-                GuestRegionMmap::new(
-                    region.mmap_region(file)?,
-                    GuestAddress(region.guest_phys_addr),
-                )
-                .map_err(|e| {
-                    VhostUserError::ReqHandlerError(io::Error::new(io::ErrorKind::Other, e))
-                })?,
-            );
+            let guest_region = GuestRegionMmap::new(
+                region.mmap_region(file)?,
+                GuestAddress(region.guest_phys_addr),
+            )
+            .map_err(|e| {
+                VhostUserError::ReqHandlerError(io::Error::new(io::ErrorKind::Other, e))
+            })?;
             mappings.push(AddrMapping {
+                #[cfg(feature = "postcopy")]
+                local_addr: guest_region.as_ptr() as u64,
                 vmm_addr: region.user_addr,
                 size: region.memory_size,
                 gpa_base: region.guest_phys_addr,
             });
+            regions.push(guest_region);
         }
 
         let mem = GuestMemoryMmap::from_regions(regions).map_err(|e| {
@@ -570,6 +581,14 @@ where
             })?,
         );
 
+        let addr_mapping = AddrMapping {
+            #[cfg(feature = "postcopy")]
+            local_addr: guest_region.as_ptr() as u64,
+            vmm_addr: region.user_addr,
+            size: region.memory_size,
+            gpa_base: region.guest_phys_addr,
+        };
+
         let mem = self
             .atomic_mem
             .memory()
@@ -586,11 +605,7 @@ where
                 VhostUserError::ReqHandlerError(io::Error::new(io::ErrorKind::Other, e))
             })?;
 
-        self.mappings.push(AddrMapping {
-            vmm_addr: region.user_addr,
-            size: region.memory_size,
-            gpa_base: region.guest_phys_addr,
-        });
+        self.mappings.push(addr_mapping);
 
         Ok(())
     }
@@ -633,6 +648,68 @@ where
         self.backend
             .check_device_state()
             .map_err(VhostUserError::ReqHandlerError)
+    }
+
+    #[cfg(feature = "postcopy")]
+    fn postcopy_advice(&mut self) -> VhostUserResult<File> {
+        let mut uffd_builder = UffdBuilder::new();
+
+        let uffd = uffd_builder
+            .close_on_exec(true)
+            .non_blocking(true)
+            .user_mode_only(false)
+            .create()
+            .map_err(|e| {
+                VhostUserError::ReqHandlerError(io::Error::new(io::ErrorKind::Other, e))
+            })?;
+
+        // We need to duplicate the uffd fd because we need both
+        // to return File with fd and store fd inside uffd.
+        //
+        // SAFETY:
+        // We know that uffd is correctly created.
+        // This means fd inside uffd is also a valid fd.
+        // Duplicating a valid fd is safe.
+        let uffd_dup = unsafe { libc::dup(uffd.as_raw_fd()) };
+        if uffd_dup < 0 {
+            return Err(VhostUserError::ReqHandlerError(io::Error::last_os_error()));
+        }
+
+        // SAFETY:
+        // We know that uffd_dup is a valid fd.
+        let uffd_file = unsafe { File::from_raw_fd(uffd_dup) };
+
+        self.uffd = Some(uffd);
+
+        Ok(uffd_file)
+    }
+
+    #[cfg(feature = "postcopy")]
+    fn postcopy_listen(&mut self) -> VhostUserResult<()> {
+        let Some(ref uffd) = self.uffd else {
+            return Err(VhostUserError::ReqHandlerError(io::Error::new(
+                io::ErrorKind::Other,
+                "No registered UFFD handler",
+            )));
+        };
+
+        for mapping in self.mappings.iter() {
+            uffd.register(
+                mapping.local_addr as *mut libc::c_void,
+                mapping.size as usize,
+            )
+            .map_err(|e| {
+                VhostUserError::ReqHandlerError(io::Error::new(io::ErrorKind::Other, e))
+            })?;
+        }
+
+        Ok(())
+    }
+
+    #[cfg(feature = "postcopy")]
+    fn postcopy_end(&mut self) -> VhostUserResult<()> {
+        self.uffd = None;
+        Ok(())
     }
 }
 
--- a/vhost-user-backend/src/lib.rs
+++ b/vhost-user-backend/src/lib.rs
@@ -33,6 +33,12 @@ pub use self::vring::{
     VringMutex, VringRwLock, VringState, VringStateGuard, VringStateMutGuard, VringT,
 };
 
+/// Due to the way `xen` handles memory mappings we can not combine it with
+/// `postcopy` feature which relies on persistent memory mappings. Thus we
+/// disallow enabling both features at the same time.
+#[cfg(all(feature = "postcopy", feature = "xen"))]
+compile_error!("Both `postcopy` and `xen` features can not be enabled at the same time.");
+
 /// An alias for `GuestMemoryAtomic<GuestMemoryMmap<B>>` to simplify code.
 type GM<B> = GuestMemoryAtomic<GuestMemoryMmap<B>>;
 
--- a/vhost-user-backend/tests/vhost-user-server.rs
+++ b/vhost-user-backend/tests/vhost-user-server.rs
@@ -292,3 +292,34 @@ fn vhost_user_get_inflight(path: &Path, barrier: Arc<Barrier>) {
 fn test_vhost_user_get_inflight() {
     vhost_user_server(vhost_user_get_inflight);
 }
+
+#[cfg(feature = "postcopy")]
+fn vhost_user_postcopy_advise(path: &Path, barrier: Arc<Barrier>) {
+    let mut frontend = setup_frontend(path, barrier);
+    let _uffd_file = frontend.postcopy_advise().unwrap();
+}
+
+#[cfg(feature = "postcopy")]
+fn vhost_user_postcopy_listen(path: &Path, barrier: Arc<Barrier>) {
+    let mut frontend = setup_frontend(path, barrier);
+    let _uffd_file = frontend.postcopy_advise().unwrap();
+    frontend.postcopy_listen().unwrap();
+}
+
+#[cfg(feature = "postcopy")]
+fn vhost_user_postcopy_end(path: &Path, barrier: Arc<Barrier>) {
+    let mut frontend = setup_frontend(path, barrier);
+    let _uffd_file = frontend.postcopy_advise().unwrap();
+    frontend.postcopy_listen().unwrap();
+    frontend.postcopy_end().unwrap();
+}
+
+// These tests need an access to the `/dev/userfaultfd`
+// in order to pass.
+#[cfg(feature = "postcopy")]
+#[test]
+fn test_vhost_user_postcopy() {
+    vhost_user_server(vhost_user_postcopy_advise);
+    vhost_user_server(vhost_user_postcopy_listen);
+    vhost_user_server(vhost_user_postcopy_end);
+}
--- a/vhost/src/lib.rs
+++ b/vhost/src/lib.rs
@@ -51,6 +51,12 @@ pub mod vhost_user;
 #[cfg(feature = "vhost-vsock")]
 pub mod vsock;
 
+/// Due to the way `xen` handles memory mappings we can not combine it with
+/// `postcopy` feature which relies on persistent memory mappings. Thus we
+/// disallow enabling both features at the same time.
+#[cfg(all(feature = "postcopy", feature = "xen"))]
+compile_error!("Both `postcopy` and `xen` features can not be enabled at the same time.");
+
 /// Error codes for vhost operations
 #[derive(Debug)]
 pub enum Error {
--- a/vhost/src/vhost_user/frontend.rs
+++ b/vhost/src/vhost_user/frontend.rs
@@ -72,6 +72,26 @@ pub trait VhostUserFrontend: VhostBackend {
 
     /// Remove a guest memory mapping from vhost.
     fn remove_mem_region(&mut self, region: &VhostUserMemoryRegionInfo) -> Result<()>;
+
+    /// Sends VHOST_USER_POSTCOPY_ADVISE msg to the backend
+    /// initiating the beginning of the postcopy process.
+    /// Backend will return a userfaultfd.
+    #[cfg(feature = "postcopy")]
+    fn postcopy_advise(&mut self) -> Result<File>;
+
+    /// Sends VHOST_USER_POSTCOPY_LISTEN msg to the backend
+    /// telling it to register its memory regions with
+    /// userfaultfd previously received through the
+    /// [`VhostUserFrontend::postcopy_advise`] call.
+    #[cfg(feature = "postcopy")]
+    fn postcopy_listen(&mut self) -> Result<()>;
+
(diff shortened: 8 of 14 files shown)


Ablu (commented):
I am not sure if this works with operation modes where regions are mapped on demand (Xen grants, for example). But I will let @vireshk comment on that.

Otherwise, I think commit messages are pretty bare-bone. Not everyone (at least not me :)) is an expert on these migration topics so it might help to explain what is happening in more detail.

Ablu:
@germag: This may be related to your work on migration?

vhost/src/vhost_user/message.rs
```diff
@@ -437,6 +437,16 @@ bitflags! {
     }
 }
 
+/// An empty message.
+#[repr(packed)]
```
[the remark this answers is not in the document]
ShadowCurse: Considering the struct is zero sized, I removed the `#[repr(..)]` as it is not needed.

vhost-user-backend/src/handler.rs
```diff
@@ -615,6 +623,53 @@ where
...
+            .close_on_exec(true)
+            .non_blocking(true)
+            .user_mode_only(false)
+            .create()
+            .map_err(|e| {
+                VhostUserError::ReqHandlerError(io::Error::new(io::ErrorKind::Other, e))
+            })?;
+
+        // SAFETY:
+        // We know that uffd is a valid fd.
```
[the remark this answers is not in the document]
ShadowCurse: Thx for noticing this. I added a step with duplication of the fd and more comments.

ShadowCurse:
Hi @Ablu, I have addressed comments you left before, please take another look. Also @vireshk and @germag, would be nice to hear your thoughts as well.

Regarding ci not passing with musl toolchain, I am working on this.

stefano-garzarella:
@ShadowCurse did you test the changes of `rust-vmm-container` locally?
You can use `./docker.sh build` to build a local image and try to build this branch through it.

For example, what I usually do is:
```
cd rust-vmm-container
./docker.sh build

cd vhost
# I'm using podman, but with docker should be similar
podman run -it --rm --security-opt=seccomp=unconfined --volume $(pwd):/workdir --workdir /workdir localhost/rustvmm/dev:v33_x86_64 
cargo build --target x86_64-unknown-linux-musl
```

vhost-user-backend/src/handler.rs
```diff
@@ -615,6 +623,53 @@ where
...
+        let Some(ref uffd) = self.uffd else {
+            return Err(VhostUserError::ReqHandlerError(io::Error::new(
+                io::ErrorKind::Other,
+                "No registered UFFD handler",
+            )));
+        };
+
+        for mapping in self.mappings.iter() {
+            uffd.register(
+                mapping.local_addr as *mut libc::c_void,
```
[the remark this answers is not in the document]
vireshk: Hey, sorry about missing this completely somehow.. So yes, I had a look at code again.. With Xen's Grant memory mapping (mapping on the fly once the buffers are available), this would lead to:

https://github.com/rust-vmm/vm-memory/blob/main/src/mmap_xen.rs#L890

which will return `null_mut()`.

The entire region isn't mapped in advance and we get to that only when we get the final request.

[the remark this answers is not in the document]
ShadowCurse: Hmm.. so how would vhost-backend even be configured with Xen mappings? It is just in this case mappings are created here https://github.com/rust-vmm/vhost/blob/321085b2ea7bd818942be7954f3daaba1fa34293/vhost-user-backend/src/handler.rs#L282. The backend implementation in this repo does not use Xen mappings, so there is should be no way for the pointer to be `null`. Am I missing smth?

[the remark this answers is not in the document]
Ablu: > The backend implementation in this repo does not use Xen mappings, so there is should be no way for the pointer to be `null`. Am I missing smth?

The `xen` feature of the crate allows enabling the Xen mapping. The vhost-device daemons (or virtiofsd) use that for Xen support. With Xen support no virtual address is known until a slice is attempted to be accessed (at which point it will be mapped only for the needed access duration).

So I think this would probably need to be integrated with vm-memory (which handles the mapping and unmapping) to some degree. Maybe the existing dirty page tracking may help there...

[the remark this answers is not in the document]
ShadowCurse: So do I understand correctly that when Xen is used with Postcopy feature, the registration for uffd should happen right after memory is mapped. And unregistered after operation is finished before unmapping? This seems like a lot of movement and overall I am not sure if anyone will use this specific combination.

What if I restrict the postcopy support to be only availabe when Xen feature is not used?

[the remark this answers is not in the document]
Ablu: > So do I understand correctly that when `xen` is used with `VHOST_USER_PROTOCOL_F_PAGEFAULT` feature, the registration for uffd should happen right after memory is mapped. And unregistered after operation is finished before unmapping? This seems like a lot of movement and overall I am not sure if anyone will use this specific combination.

I am not sure whether that is what should be happening, I only pointed out that that is the only period over which the mapping guaranteed to be known.

> What if I restrict the postcopy support to be only availabe when `xen` feature is not used?

IMHO it is not so much a Xen question, but more of an architecture question. Xen (and I was told pKVM too) need "on-demand" mapping. Hence, we deprecated other places that allowed direct pointer access without mapping guards in place. But I will have to leave it to someone more familiar with the technology to comment on how this would be best modeled in an hypervisor-agnostic way.

ShadowCurse:
@Ablu I have updated the PR to disable `postcopy` messages if `xen` feature is enabled. I think this should work for the majority of use cases, while `xen` + `postcopy` can be added later if need arises.

Ablu:
> @Ablu I have updated the PR to disable `postcopy` messages if `xen` feature is enabled. I think this should work for the majority of use cases, while `xen` + `postcopy` can be added later if need arises.

Hm. I must admit that I am not particular thrilled to add further Xen vs non-Xen flags. In fact, I would prefer this way of getting access to virtual addresses altogether as it seems to be confusing.

But I am happy to hear other peoples opinions.

vhost-user-backend/src/handler.rs
```diff
@@ -617,6 +632,68 @@ where
...
+                VhostUserError::ReqHandlerError(io::Error::new(io::ErrorKind::Other, e))
+            })?;
+        }
+
+        Ok(())
+    }
+
+    #[cfg(not(feature = "xen"))]
+    fn postcopy_end(&mut self) -> VhostUserResult<()> {
+        _ = self.uffd.take();
```
stefano-garzarella, line 694: If we are not using the return value, what about using `self.uffd = None` ?

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 75.0,
+  "coverage_score": 73.30,
```
stefano-garzarella, line 2: Can we try to call the new API in `vhost-user-backend/src/handler.rs` or it is too complicated?
If it is the case, can you explain it in the commit description?

stefano-garzarella (changes requested):

vhost/src/vhost_user/frontend.rs
```diff
@@ -72,6 +72,19 @@ pub trait VhostUserFrontend: VhostBackend {
...
+    #[cfg(not(feature = "xen"))]
+    fn vhost_user_postcopy_advise(&mut self) -> Result<File>;
+
+    /// Sends VHOST_USER_POSTCOPY_LISTEN msg to the backend.
+    #[cfg(not(feature = "xen"))]
+    fn vhost_user_postcopy_listen(&mut self) -> Result<()>;
+
+    /// Sends VHOST_USER_POSTCOPY_END msg to the backend.
+    #[cfg(not(feature = "xen"))]
+    fn vhost_user_postcopy_end(&mut self) -> Result<()>;
```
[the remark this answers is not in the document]
ShadowCurse: fixed names and updated the comments.

vhost-user-backend/src/handler.rs
ShadowCurse: I think the `Option::take` is more idiomatic. Basically we assign whatever is in the `uffd` to `_` and it will be dropped at the end of the scope.

coverage_config_x86_64.json
ShadowCurse: The whole `handler.rs` is not tested and I assume this is because it is too difficult to create the handler for the test. I will update the commit message to mention this.

stefano-garzarella:
nit: I suggest to use imperative form in commit (title and description). https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#describe-your-changes . And possibly I'd use `vhost:` and `vhost-user-backend:` prefix in the title since we have multiple crates in this workspace and that allows us to better understand the git story.

vhost-user-backend/src/handler.rs
stefano-garzarella: IMHO it's clearer by assigning `None` that we're basically dropping whatever value was there before. It might make sense to use `take()` if we then check that there really was something there (should we do that and return an error?).

coverage_config_x86_64.json
stefano-garzarella: Sorry, I posted the wrong path, I meant if we can add the test here: `vhost-user-backend/tests/vhost-user-server.rs` , but since the backend is the dummy one, I'm not sure how much we can increase the coverage.

ShadowCurse:
> nit: I suggest to use imperative form in commit (title and description). https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#describe-your-changes . And possibly I'd use `vhost:` and `vhost-user-backend:` prefix in the title since we have multiple crates in this workspace and that allows us to better understand the git story.

Updated commits. Do they look better now?

vhost/src/vhost_user/frontend.rs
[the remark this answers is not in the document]
ShadowCurse: This is a typo, thx noticing. Fixed.

vhost-user-backend/src/handler.rs
ShadowCurse: Ok, switched to `self.uffd = None`. I don't think there should be an error if uffd does not exist. Qemu does not error out in this case as well.

coverage_config_x86_64.json
ShadowCurse: The `MockVhostBackend` does not use `VhostUserHandler`, so it can not be tested without rewriting the `MockVhostBackend` and all of the tests.

stefano-garzarella: Okay, thanks for checking!

vhost-user-backend/src/handler.rs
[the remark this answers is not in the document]
stefano-garzarella: @Ablu @vireshk I don't know Xen at all, so what do you think about the current implementation (postcopy supported only when `xen` feature is not enabled) ?

stefano-garzarella:
> Updated commits. Do they look better now?

Yep, thanks!
I'd avoided "can" and used "support", for example: "vhost/frontend: support POSTCOPY messages", but I don't want to be too pedantic, it's fine even as it is now ;-)

BTW the code LGTM, but I'd wait a bit a comment from Viresh and Erik.

vhost-user-backend/Cargo.toml
```diff
@@ -9,11 +9,13 @@ edition = "2021"
 license = "Apache-2.0"
 
 [features]
+default = ["userfaultfd"]
 xen = ["vm-memory/xen", "vhost/xen"]
 
 [dependencies]
 libc = "0.2.39"
 log = "0.4.17"
+userfaultfd = { version = "0.7.0", optional = true }
```
stefano-garzarella, line 18: Just another question (more for the user of this crate, like `virtiofsd` and `vhost-device`). Since we are adding a new dependency, and maybe some backends don't need the `postcopy` feature, should we gate this with a new feature (enabled by default)?

Just to avoid some unneeded dependency during packaging.

@germag @vireshk @epilys @stsquad @Ablu WDYT?

ShadowCurse: As of now `userfaultfd` crate is optional and enabled by default. So if users don't want the unneeded dependency they can specify the crate with `vhost = { ..., default-features=false }`. 

Or I can switch to gate postcopy with separate feature which is not enabled by default (so it will be like `xen` feature). Then I can update 
```rust
#[cfg(not(feature = "xen"))]
```
 into 
```rust
#[cfg(all(feautre = "postcopy", not(feature = "xen")))]
```
Hm, I think second option should be the best here.

stefano-garzarella: > As of now `userfaultfd` crate is optional and enabled by default. So if users don't want the unneeded dependency they can specify the crate with `vhost = { ..., default-features=false }`.
> 

Right, but what happens if in the future we will add another feature, and we want to enable only `postcopy`?

> Or I can switch to gate postcopy with separate feature which is not enabled by default (so it will be like `xen` feature). Then I can update
> 
> ```rust
> #[cfg(not(feature = "xen"))]
> ```
> 
> into
> 
> ```rust
> #[cfg(all(feautre = "postcopy", not(feature = "xen")))]
> ```
> 
> Hm, I think second option should be the best here.

Yeah, I think so. Or have some kind of build failure if both are enabled.

stefano-garzarella: > Yeah, I think so. Or have some kind of build failure if both are enabled.

Here there is an example: https://doc.rust-lang.org/std/macro.compile_error.html

ShadowCurse: Ok, I updated the code, but now the CI is in the way. It runs tests and clippy with `--all-features` so it fail now.

germag: > Ok, I updated the code, but now the CI is in the way. It runs tests and clippy with `--all-features` so it fail now.

That should be fixed, because testing with `--all-features` i'st effectively _only_ testing the 'xen' feature, since that feature disables the 'default' feature in vm-memory(*)

(*)  vm-memory should be also fixed, because it allows to compile it with '--all-fetures' and then failing in runtime

ShadowCurse: Do I understand correctly that you are saying that `rust-vmm-ci` and `vm-memory` should be fixed? (The 'That' part of the first sentence is confusing) Otherwise I don's see how this crate can do a work around for this issue.

stefano-garzarella: > Do I understand correctly that you are saying that `rust-vmm-ci` and `vm-memory` should be fixed? (The 'That' part of the first sentence is confusing) Otherwise I don's see how this crate can do a work around for this issue.

yep, in `rust-vmm-ci`. One workaround we can do is to define our custom `.buildkite` pipeline.
But this is for sure for another PR. I can do that, but tomorrow I'll be off the whole day, so I can't do before Friday.

Ablu: > Just another question (more for the user of this crate, like `virtiofsd` and `vhost-device`). Since we are adding a new dependency, and maybe some backends don't need the `postcopy` feature, should we gate this with a new feature (enabled by default)?

I got no opinion here. I assume userfaultfd is just a shallow wrapper around fd operations? So I assume it does not blow up the code size by much. So I may also be fine with keeping it if adding knobs becomes a maintenance burden too.

vireshk:
> Hi @Ablu, I have addressed comments you left before, please take another look. Also @vireshk and @germag, would be nice to hear your thoughts as well.

Hey, sorry for being late on this. Yeah it is fine for now to disable the feature for Xen builds. That's fine. Thanks.

vhost-user-backend/Cargo.toml
stefano-garzarella: @Ablu yes, I see your point, but it definitely carries other dependencies, and when we go packaging (for example in fedora), it's always better to have as few dependencies as possible, especially if we know we're not going to use that feature.

Also, in this case we need gates for that code anyway, so it seems free to me.

stefano-garzarella: I posted this: https://github.com/rust-vmm/vhost/pull/229

@ShadowCurse can you rebase on it, and add `postcopy` feature in `.buildkite/rust-vmm-ci-tests.json` in `*-all-without-xen` tests?

`rust-vmm-ci` pipeline is still active, so it should fail, but `custom-rust-vmm-ci` should work.
I'll deactivate it when we merge #229

ShadowCurse: Done

stefano-garzarella: thanks! It looks like it is working good! I'll rebuild the pipeline when I'll disable the old pipeline

vhost-user-backend/Cargo.toml
```diff
@@ -10,10 +10,12 @@ license = "Apache-2.0"
 
 [features]
 xen = ["vm-memory/xen", "vhost/xen"]
+postcopy = ["vhost/postcopy", "userfaultfd"]
```
stefano-garzarella, line 13: IMHO we can enable this by default.
Do we really need the feature in `vhost` too? At the end we added just some wrappers right?

ShadowCurse: For `vhost-user-backend` I think having it as opt-in rather then default makes sense. This is quite specific protocol part + requires additional dependency.
For `vhost`, I guess we can remove feature gate and nothing will break because changes are additive. But then we have dis balance between `vhost` and `vhost-user-backend` crates.

stefano-garzarella: > For `vhost-user-backend` I think having it as opt-in rather then default makes sense. This is quite specific protocol part + requires additional dependency.

Ack, makes sense.

> For `vhost`, I guess we can remove feature gate and nothing will break because changes are additive. But then we have dis balance between `vhost` and `vhost-user-backend` crates.

I don't have a strong opinion, so I'm fine also to have it in `vhost`.

I just noticed that `coverage` now increased. I think because previously the new code was enabled by default and now it is not. Eventually we need to enable the feature in `crate_features` in `coverage_config_x86_64.json`, but if we don't know how to test it, maybe just adjust the coverage value for now.

ShadowCurse: I will adjust coverage after all rebasing is done.

vhost-user-backend/Cargo.toml
JonathanWoollett-Light: Looking at this I think there are 3 approaches we can use to fix this:

1. In vhost wrap `lib.rs` in `cfg(not(all(feature="xen", feature="postcopy")))` so when compiling with `xen` and `postcopy` it compiles an empty crate where `cargo build` and `cargo test` will work.
2. Update rust-vmm-ci to not use `--all-features` so it only tests the default features by default and crates can add additional tests for their feature sets.
3. Do https://github.com/rust-vmm/rust-vmm-ci/issues/152

I think solutions 1 and 2 are both viable and reasonably easy to implement. Both 2 and 3 will affect other crates. I think 3 raises a few too many concerns about how much it increase runtime of tests. I would probably recommend going with 1 for now as its the easiest to implement since it doesn't require external changes.

stefano-garzarella: > Looking at this I think there are 3 approaches we can use to fix this:

@JonathanWoollett-Light thanks for your suggestions!
I proposed #229 as another approach, maybe we should continue the discussion there.

> 
>     1. In vhost wrap `lib.rs` in `cfg(not(all(feature="xen", feature="postcopy")))` so when compiling with `xen` and `postcopy` it compiles an empty crate where `cargo build` and `cargo test` will work.

Does this mean that our CI will always go to test an empty crate?

> 
>     2. Update rust-vmm-ci to not use `--all-features` so it only tests the default features by default and crates can add additional tests for their feature sets.

We have PRs (like this one for example) where almost all the code is behind a feature and so CI would not run `clippy` for example on this new code. I don't know if that's good.

> 
>     3. Do [Add `cargo-all-features` rust-vmm-ci#152](https://github.com/rust-vmm/rust-vmm-ci/issues/152)

IMHO we should move toward this direction, though, somewhat limiting the testing matrix and providing only a set of features to test. So using it more to avoid testing unsupported combinations.

germag: > Looking at this I think there are 3 approaches we can use to fix this:
> 
>     1. In vhost wrap `lib.rs` in `cfg(not(all(feature="xen", feature="postcopy")))` so when compiling with `xen` and `postcopy` it compiles an empty crate where `cargo build` and `cargo test` will work.
> 
>     2. Update rust-vmm-ci to not use `--all-features` so it only tests the default features by default and crates can add additional tests for their feature sets.
> 
>     3. Do [Add `cargo-all-features` rust-vmm-ci#152](https://github.com/rust-vmm/rust-vmm-ci/issues/152)
> 
> 
> I think solutions 1 and 2 are both viable and reasonably easy to implement. Both 2 and 3 will affect other crates. I think 3 raises a few too many concerns about how much it increase runtime of tests. I would probably recommend going with 1 for now as its the easiest to implement since it doesn't require external changes.

I disagree with 1, silently creating an empty crate it's not better than current situation. Enabling xen with an incompatible feature must be a compilation error, instead of a runtime one. This is also a problem for some distributions that recommend to build crates with `--all-features` (like Arch)

stefano-garzarella: > > ```
> > 2. Update rust-vmm-ci to not use `--all-features` so it only tests the default features by default and crates can add additional tests for their feature sets.
> > ```
> 
> We have PRs (like this one for example) where almost all the code is behind a feature and so CI would not run `clippy` for example on this new code. I don't know if that's good.

Okay, I had misinterpreted, yes I think this is a viable option, then we have to do something like #229 (with just some tests included there).

vhost-user-backend/Cargo.toml
stefano-garzarella: @ShadowCurse FYI #229 is now merged

.buildkite/rust-vmm-ci-tests.json
```diff
@@ -62,7 +62,7 @@
     },
     {
       "test_name": "clippy-all-without-xen",
-      "command": "cargo clippy --workspace --bins --examples --benches --all-targets --no-default-features --features test-utils,vhost-vsock,vhost-kern,vhost-vdpa,vhost-net,vhost-user,vhost-user-frontend,vhost-user-backend -- -D warnings -D clippy::undocumented_unsafe_blocks",
+      "command": "cargo clippy --workspace --bins --examples --benches --all-targets --no-default-features --features test-utils,vhost-vsock,vhost-kern,vhost-vdpa,vhost-net,vhost-user,vhost-user-frontend,vhost-user-backend,postcopy -- -D warnings -D clippy::undocumented_unsafe_blocks",
```
stefano-garzarella, line 65: Can you add `postcopy` also at line 81 for the `check-warnings-all-without-xen` test?

vhost/src/lib.rs
```diff
@@ -51,6 +51,9 @@ pub mod vhost_user;
 #[cfg(feature = "vhost-vsock")]
 pub mod vsock;
 
+#[cfg(all(feature = "postcopy", feature = "xen"))]
+compile_error!("Both `postcopy` and `xen` features can not be enabled at the same time.");
```
stefano-garzarella, line 55: I'd document this in the README (in the `Xen support` section), explaining also why.
Maybe we can add also a little comment here with the reason behind this incompatibility.

vhost-user-backend/src/lib.rs
```diff
@@ -33,6 +33,9 @@ pub use self::vring::{
     VringMutex, VringRwLock, VringState, VringStateGuard, VringStateMutGuard, VringT,
 };
 
+#[cfg(all(feature = "postcopy", feature = "xen"))]
+compile_error!("Both `postcopy` and `xen` features can not be enabled at the same time.");
```
stefano-garzarella, line 37: Ditto (document in the README and maybe a little comment here).

.buildkite/rust-vmm-ci-tests.json
ShadowCurse: done

vhost/src/lib.rs
ShadowCurse: Added new postcopy section with explanation. Also added comments above the compile check.

vhost-user-backend/src/lib.rs
ShadowCurse: Same as above

stefano-garzarella (dismissed):

vhost-user-backend/src/handler.rs
```diff
@@ -6,10 +6,14 @@
 use std::error;
 use std::fs::File;
 use std::io;
+#[cfg(feautre = "postcopy")]
```
Ablu, line 9: Shouldn't this be `feature`? :thinking: Shouldn't this just be a no-op as it is?

Ablu (commented):
The typo makes me a bit nervous. It probably results in a no-op for all the builds, or do I miss something here? 

Could you elaborate how you tested this? Would it maybe make sense to add automated tests?

vhost-user-backend/src/handler.rs
stefano-garzarella: Great catch!

I think we didn't realize it, because we don't have any tests where we call these new APIs.

Maybe we should add a negative test (expecting an error if our `MockVhostBackend` doesn't implement them) in `vhost-user-backend/tests/vhost-user-server.rs`. That way we can at least test the API and the feature.

ShadowCurse: Thx for catching this. Fixed now. With the type the `cfg` just disables whatever is under it (so behaves same as if feature is not enabled).
I did manual testing with Firecracker where I replaced this crate with my fork and triggered all APIs from there for qemu backend.
`MockVhostBackend` in the crate will not be able to test new APIs, because it is the handler that implement them. I don's see `MockVhostBackend` creating handler and testing anything with it.

stefano-garzarella: > `MockVhostBackend` in the crate will not be able to test new APIs, because it is the handler that implement them. I don's see `MockVhostBackend` creating handler and testing anything with it.

Why you need to create a handler, the vhost-user-backend is handling it, the specific device doesn't need to do anything, no?
Isn't this similar to `add_mem_region()` for example?

We call it in the test, but `MockVhostBackend` doesn't implement any callback, since it is handled by `BackendReqHandler::handle_request`.

So, I mean something like this (Note: needs to play with features, etc.):

```diff
diff --git a/vhost-user-backend/tests/vhost-user-server.rs b/vhost-user-backend/tests/vhost-user-server.rs
index d8be86d..caa31c1 100644
--- a/vhost-user-backend/tests/vhost-user-server.rs
+++ b/vhost-user-backend/tests/vhost-user-server.rs
@@ -292,3 +292,14 @@ fn vhost_user_get_inflight(path: &Path, barrier: Arc<Barrier>) {
 fn test_vhost_user_get_inflight() {
     vhost_user_server(vhost_user_get_inflight);
 }
+
+fn vhost_user_postcopy(path: &Path, barrier: Arc<Barrier>) {
+    let mut frontend = setup_frontend(path, barrier);
+
+    frontend.postcopy_end().unwrap();
+}
+
+#[test]
+fn test_vhost_user_postcopy() {
+    vhost_user_server(vhost_user_postcopy);
+}
```

It is working, while `postcopy_advice` return an error, I didn't have time to investigate, maybe I need some special permission to create the `userfaultfd`.

ShadowCurse: Oh ok. Sorry, somehow I totally skipped this layered cake of abstractions used in the tests. Would have never thought that in the `vhost-user-backend/tests` there will be tests for both crates.
Anyway I added tests as you suggested. In order to use `uffd` process need access to `dev/userfaultfd`. Considering tests do pass, it seems they run as root in the container, so everything is fine.

vhost-user-backend/tests/vhost-user-server.rs
```diff
@@ -292,3 +292,32 @@ fn vhost_user_get_inflight(path: &Path, barrier: Arc<Barrier>) {
...
+fn vhost_user_postcopy_end(path: &Path, barrier: Arc<Barrier>) {
+    let mut frontend = setup_frontend(path, barrier);
+    let _uffd_file = frontend.postcopy_advise().unwrap();
+    frontend.postcopy_listen().unwrap();
+    frontend.postcopy_end().unwrap();
+}
+
+#[cfg(feature = "postcopy")]
+#[test]
+fn test_vhost_user_postcopy() {
```
stefano-garzarella, line 319: Can you add a comment here about `/dev/userfaultfd` permission?

That way if someone launches the tests locally, they immediately understand the problem if it fails.

vhost-user-backend/src/handler.rs
stefano-garzarella: > Oh ok. Sorry, somehow I totally skipped this layered cake of abstractions used in the tests. Would have never thought that in the `vhost-user-backend/tests` there will be tests for both crates.

The two crates are very connected to each other (so we moved them to the same workspace), especially to test the backend, we need the frontend to send messages. Since in the end our frontend is just a message wrapper, we use it to test the backend.

stefano-garzarella:
Now that we have tests for the new code, maybe we can enable `postcopy` for coverage:

```diff

diff --git a/coverage_config_x86_64.json b/coverage_config_x86_64.json
index d201e56..9923e17 100644
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
   "coverage_score": 75.0,
   "exclude_path": "vhost/src/vhost_kern/",
-  "crate_features": "vhost/vhost-user-frontend,vhost/vhost-user-backend"
+  "crate_features": "vhost/vhost-user-frontend,vhost/vhost-user-backend,postcopy"
 }
```

I tried it locally and the coverage test passes successfully.

vhost-user-backend/tests/vhost-user-server.rs
ShadowCurse: Added a small comment here. Also added more info in the `vhost-user-backend/README.md`.

vhost-user-backend/src/handler.rs
ShadowCurse: Now it is starting to make some sense, but the amount of indirections and abstractions makes it hard to comprehend what is going on here.

stefano-garzarella: yeah, I know :-(

stefano-garzarella (dismissed):
LGTM!

Nit: there is a little typo in the last commit title (s/coverade/coverage).

Ablu (approved):

Ablu:
Thanks!

PR #224, opened 2024-01-09
vhost-user-backend: bump up MAX_MEM_SLOTS to 509

germag:
### Summary of the PR

Let's support up to 509 memslots, just like vhost in the kernel usually does. This is required to properly support memory hotplug, either using multyiple DIMMs (ACPI supports up to 256) or using virtio-mem.

The 509 used to be the KVM limit, it supported 512, but 3 were used for internal purposes. Currently, KVM supports more than 512, but it usually doesn't make use of more than ~260 (i.e., 256 DIMMs + boot memory), except when other memory devices like PCI devices with BARs are used. So, 509 seems to work well for vhost in the kernel.

Details can be found in the QEMU change that made virtio-mem consume up to 256 memslots accross all virtio-mem devices. [1]

509 memslots implies 509 VMAs/mappings in the worst case (even though, in practice with virtio-mem we won't be seeing more than ~260 in most setups).

With max_map_count under Linux defaulting to 64k, 509 memslots still correspond to less than 1% of the maxium number of mappings. There are plenty left for the application to consume.

[1] https://lore.kernel.org/all/20230926185738.277351-1-david@redhat.com/

### DIFF
--- a/vhost-user-backend/src/handler.rs
+++ b/vhost-user-backend/src/handler.rs
@@ -30,7 +30,10 @@ use super::event_loop::{VringEpollError, VringEpollResult};
 use super::vring::VringT;
 use super::GM;
 
-const MAX_MEM_SLOTS: u64 = 32;
+// vhost in the kernel usually supports 509 mem slots.
+// The 509 used to be the KVM limit, it supported 512, but 3 were used
+// for internal purposes (nowadays, it supports more than that).
+const MAX_MEM_SLOTS: u64 = 509;
 
 #[derive(Debug)]
 /// Errors related to vhost-user handler.
--- a/vhost-user-backend/tests/vhost-user-server.rs
+++ b/vhost-user-backend/tests/vhost-user-server.rs
@@ -211,7 +211,7 @@ fn vhost_user_client(path: &Path, barrier: Arc<Barrier>) {
     let state = frontend.get_vring_base(0).unwrap();
     frontend.set_vring_base(0, state as u16).unwrap();
 
-    assert_eq!(frontend.get_max_mem_slots().unwrap(), 32);
+    assert_eq!(frontend.get_max_mem_slots().unwrap(), 509);
     let region = VhostUserMemoryRegionInfo::new(0x800000, 0x100000, addr, 0, fd.file().as_raw_fd());
     frontend.add_mem_region(&region).unwrap();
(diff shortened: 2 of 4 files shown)


vhost-user-backend/src/handler.rs
```diff
@@ -30,7 +30,7 @@ use super::event_loop::{VringEpollError, VringEpollResult};
 use super::vring::VringT;
 use super::GM;
 
-const MAX_MEM_SLOTS: u64 = 32;
+const MAX_MEM_SLOTS: u64 = 509;
```
stefano-garzarella, line 33: The changes make sense to me, and being aligned with `vhost` kernel module as well.
How about adding a comment here to shortly explain why we use this value?

germag:
v2:
- Add a short explanation about the 509 "magic number"
- Fix commit message typos

stefano-garzarella (approved):

Ablu (approved):
LGTM. But I will wait with merging to give @sboeuf a chance to comment (since he added the code)

sboeuf (approved):

stefano-garzarella:
It looks like this issue prevents any vhost-user device to work well with virtio-mem.
Here an issue with virtiofsd and virtio-mem: https://issues.redhat.com/browse/RHEL-15317

> 
> We want to support dynamically using multiple memslots to expose virtio-mem device memory to the VM; using dynamically multiple memslots can drastically reduce memory overhead in the hypervisor (especially, KVM), when a device exposes only comparatively small memory towards the VM, compared to its possible maximum size.
> 
> For QEMU, the feature is enabled using "dynamic-memslots=on". With "dynamic-memslots=off", the feature is disabled and we default to using a single large memslot statically.
> 
> **In combination with vhost devices, this new feature can be problematic if the devices support less than 509 memslots.** If such devices are created before the virtio-mem device in QEMU, virtio-mem will default to the old handling of using a single memslot only. If the devices are created after the virtio-mem devices (on the cmdline, hotplug of such devices), QEMU will bail out.

@sboeuf @Ablu WDYT about a new `vhost-user-backend` release (i.e. v0.13.0)?

Ablu:
@stefano-garzarella: I do not know the mechanics well enough to have an opinion here. Feel free to move ahead with a release.

PR #229, opened 2024-01-19
tests: add rust-vmm-ci-tests.json pipeline

stefano-garzarella:
### Summary of the PR

This is based on `rust-vmm-ci/.buildkite/test_description.json`.
We can't run rust-vmm-ci tests because they enable all the features with
`--all-features` and our crates have features that may not be compatible with
others (e.g. `xen`). Waiting to solve this problem in rust-vmm-ci (see
https://github.com/rust-vmm/rust-vmm-ci/issues/152), we use a custom
pipeline based on that but that does not use `--all-features`.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/.buildkite/README.md
+++ b/.buildkite/README.md
@@ -1,0 +1,10 @@
+# buildkite custom pipelines
+
+This folder contains the custom pipelines for this repository.
+
+If we add a new pipeline we need to enable it in
+https://buildkite.com/rust-vmm/vhost-ci/steps
+
+Custom pipelines currently defined are:
+- `custom-tests.json`
+  Custom tests to enable only certain features.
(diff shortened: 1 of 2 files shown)


Ablu (dismissed):

stefano-garzarella:
v2:
- Shortened the test names because they are needed in DOCKER_PLUGIN_CONFIG for some configurations

stefano-garzarella:
For now I have left the `rust-vmm-ci` pipeline active, but as soon as we merge this one I will deactivate it.

stefano-garzarella:
Should we run `clippy` and `cargo test` also with default features? Or `cargo build` is enough?
(for now I just tried to replicate `rust-vmm-ci`, so maybe we can think about that later)

Ablu:
> Should we run `clippy` and `cargo test` also with default features? Or `cargo build` is enough?
> (for now I just tried to replicate `rust-vmm-ci`, so maybe we can think about that later)

I do not mind. So let's keep it simple? Probably time is better spent on trying to solve https://github.com/rust-vmm/rust-vmm-ci/issues/152 :).

Ablu (approved):

PR #238, opened 2024-04-11
vhost-user-backend: Prepare v0.14.1 release

MatiasVara:
### Summary of the PR

This release changes the dependency for virtio-queue to 0.12.0.

### Requirements

requirements:

  test.


slp (dismissed):
LGTM, thanks.

Ablu:
Since virtio_queue::Queue appears in public API of vhost (for example `VringState::get_queue()`), it is a public dependency. So I think it should be a major bump of vhost-user-backend?

stefano-garzarella:
> Since `virtio_queue::Queue` appears in public API of vhost (for example `VringState::get_queue()`), it is a public dependency. So I think it should be a major bump of vhost-user-backend?

I had the same doubt. I usually launch `cargo semver-checks` to check, but I'm in Korea without a laptop, so I can't :)

https://crates.io/crates/cargo-semver-checks

Ablu:
> I had the same doubt. I usually launch `cargo semver-checks` to check, but I'm in Korea without a laptop, so I can't :)
> 
> https://crates.io/crates/cargo-semver-checks

It would be great if that could pick it up, but unfortunately, it currently does not: https://github.com/obi1kenobi/cargo-semver-checks/issues/447 :(.

MatiasVara:
> Since `virtio_queue::Queue` appears in public API of vhost (for example `VringState::get_queue()`), it is a public dependency. So I think it should be a major bump of vhost-user-backend?

Thanks Eric for the comment. Is there any document that specifies when a major release is required? If it is a major, the release would be `0.15.0`?

Thanks.

Ablu:
> Thanks Eric for the comment. Is there any document that specifies when a major release is required?

There is https://doc.rust-lang.org/cargo/reference/semver.html for general documentation, but in the end detail questions turn into an analysis whether a "breaking change" may happen (and what "breaking" means)... So we could analyse that in great detail to understand all the edge-cases, or we just bump the major if in doubt :).

> If it is a major, the release would be `0.15.0`?

Rust/Cargo considers the first non-zero number as "major" so yes, `0.15.0` would be the next major version.

Ablu (approved):

stefano-garzarella:
@MatiasVara @Ablu @slp thanks for the release! I see it in creates.io. I just pushed the tag and created the release here, please double-check ;-)

Issue #236, opened 2024-03-12
Support for VHOST_USER_GPU_SET_SOCKET, and the protocol on this socket


mtjhrc:
Hi,
me and @dorindabassey are working on a vhost-user-gpu implementation, see: https://github.com/rust-vmm/vhost-device/issues/598.

We are aiming at being compatible with QEMU's displays, which means supporting the `VHOST_USER_GPU_SET_SOCKET` message and the protocol on this socket. 
For specification see: https://www.qemu.org/docs/master/interop/vhost-user-gpu.html.

The `VHOST_USER_GPU_SET_SOCKET` is currently defined in the `FrontendReq` enum:
 https://github.com/rust-vmm/vhost/blob/6ce9d36b1e7b84e53ece51acd87e5cd8f65ef12c/vhost/src/vhost_user/message.rs#L162) 

but is not being handled here:
https://github.com/rust-vmm/vhost/blob/6ce9d36b1e7b84e53ece51acd87e5cd8f65ef12c/vhost/src/vhost_user/backend_req_handler.rs#L370-L376

The protocol on this socket is nearly the same as the normal vhost user protocol and the header structure is exactly the same as the existing `VhostUserMsgHeader`. The problem are the flags, because only the REPLY flag is specified, not even a VERSION flag is used.

I would like to implement this protocol, but my main problem is I don't know if adding new device specific protocol code into this project is ok. I see that there are FS specific things, but they might be removed (see https://github.com/rust-vmm/vhost/issues/213)

I can think of 3 solution with diferent levels genericness:
- creating a `GpuBackend` struct and passing it to a `set_gpu_socket` when `VHOST_USER_GPU_SET_SOCKET` is received in a similar fashion to the existing `Backend` and `set_backend_req_fd` callback
 https://github.com/rust-vmm/vhost/blob/6ce9d36b1e7b84e53ece51acd87e5cd8f65ef12c/vhost/src/vhost_user/backend_req_handler.rs#L535-L540
- having a `set_gpu_socket` but passing it a `UnixStream` and let the user parse the messages
  - here it would be nice to make some utilities public from vhost-user, such as `Endpoint` so there is no code duplication
- not handling `VHOST_USER_GPU_SET_SOCKET` in this project at all, but providing a fully generic way to handle custom messages from the frontend, for messages that are not handled by the `handle_request` referenced above.

I am not sure on what is the stance on device-specific and/or non-standard features in this project, any suggestions?

Ablu:
[...]

PR #239, opened 2024-04-18
Add support for VHOST_USER_GPU_SET_SOCKET

mtjhrc:
### Summary of the PR
Me and @dorindabassey have been working on this PR, which adds support for the QEMU's [vhost-user-gpu protocol](https://www.qemu.org/docs/master/interop/vhost-user-gpu.html). 
This is a protocol for sending messeges from backend to the fronted, it is needed to implement display output in QEMU to implement a [vhost-user-gpu device](https://github.com/rust-vmm/vhost-device/issues/598).
See issue #236 

This feature is very similar to the `SET_BACKEND_REQ_FD`, so it is implemented in a similar way.
Support is only added for sending these vhost messages from backend to frontend. Support for handling this protocol on frontend is not implemented.

The protocol is very similar to the normal vhost protocol, but the valid header flags of the `VhostUserMsgHeader` and the newly introduced `VhostUserGpuMsgHeader` are different. (There is no version bit, only reply bit is specified). So the `Endpoint` utility was made generic over the header type, not just the request. Unfortunately there is another issue, that the messages `VHOST_USER_GPU_SCANOUT` and `VHOST_USER_GPU_CURSOR_UPDATE` are bigger than the `MAX_MSG_SIZE`. So we added methods for sending bigger messages to `Endpoint`.
https://github.com/rust-vmm/vhost/blob/bd6bf1325b4930d4123f6154ce755bd5289c617b/vhost/src/vhost_user/message.rs#L26-L33

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-user-backend/Cargo.toml
+++ b/vhost-user-backend/Cargo.toml
@@ -11,6 +11,7 @@ license = "Apache-2.0"
 [features]
 xen = ["vm-memory/xen", "vhost/xen"]
 postcopy = ["vhost/postcopy", "userfaultfd"]
+gpu-socket = ["vhost/gpu-socket"]
 
 [dependencies]
 libc = "0.2.39"
--- a/vhost-user-backend/src/backend.rs
+++ b/vhost-user-backend/src/backend.rs
@@ -31,6 +31,9 @@ use vm_memory::bitmap::Bitmap;
 use vmm_sys_util::epoll::EventSet;
 use vmm_sys_util::eventfd::EventFd;
 
+#[cfg(feature = "gpu-socket")]
+use vhost::vhost_user::GpuBackend;
+
 use super::vring::VringT;
 use super::GM;
 
@@ -83,6 +86,14 @@ pub trait VhostUserBackend: Send + Sync {
     /// A default implementation is provided as we cannot expect all backends to implement this
     /// function.
     fn set_backend_req_fd(&self, _backend: Backend) {}
+
+    #[cfg(feature = "gpu-socket")]
+    /// Set handler for communicating with the frontend by the gpu specific backend communication
+    /// channel.
+    ///
+    /// This method only exits when the crate feature gpu-socket is enabled, because this is only
+    /// useful for a gpu device.
+    fn set_gpu_socket(&self, _gpu_backend: GpuBackend);
 
     /// Get the map to map queue index to worker thread index.
     ///
@@ -194,6 +205,14 @@ pub trait VhostUserBackendMut: Send + Sync {
     /// function.
     fn set_backend_req_fd(&mut self, _backend: Backend) {}
 
+    #[cfg(feature = "gpu-socket")]
+    /// Set handler for communicating with the frontend by the gpu specific backend communication
+    /// channel.
+    ///
+    /// This method only exits when the crate feature gpu-socket is enabled, because this is only
+    /// useful for a gpu device.
+    fn set_gpu_socket(&mut self, gpu_backend: GpuBackend);
+
     /// Get the map to map queue index to worker thread index.
     ///
     /// A return value of [2, 2, 4] means: the first two queues will be handled by worker thread 0,
@@ -299,6 +318,11 @@ impl<T: VhostUserBackend> VhostUserBackend for Arc<T> {
         self.deref().set_backend_req_fd(backend)
     }
 
+    #[cfg(feature = "gpu-socket")]
+    fn set_gpu_socket(&self, gpu_backend: GpuBackend) {
+        self.deref().set_gpu_socket(gpu_backend)
+    }
+
     fn queues_per_thread(&self) -> Vec<u64> {
         self.deref().queues_per_thread()
     }
@@ -374,6 +398,11 @@ impl<T: VhostUserBackendMut> VhostUserBackend for Mutex<T> {
 
     fn set_backend_req_fd(&self, backend: Backend) {
         self.lock().unwrap().set_backend_req_fd(backend)
+    }
+
+    #[cfg(feature = "gpu-socket")]
+    fn set_gpu_socket(&self, gpu_backend: GpuBackend) {
+        self.lock().unwrap().set_gpu_socket(gpu_backend)
     }
 
     fn queues_per_thread(&self) -> Vec<u64> {
@@ -454,6 +483,11 @@ impl<T: VhostUserBackendMut> VhostUserBackend for RwLock<T> {
 
     fn set_backend_req_fd(&self, backend: Backend) {
         self.write().unwrap().set_backend_req_fd(backend)
+    }
+
+    #[cfg(feature = "gpu-socket")]
+    fn set_gpu_socket(&self, gpu_backend: GpuBackend) {
+        self.write().unwrap().set_gpu_socket(gpu_backend)
     }
 
     fn queues_per_thread(&self) -> Vec<u64> {
@@ -575,6 +609,9 @@ pub mod tests {
         }
 
         fn set_backend_req_fd(&mut self, _backend: Backend) {}
+
+        #[cfg(feature = "gpu-socket")]
+        fn set_gpu_socket(&mut self, _gpu_backend: GpuBackend) {}
 
         fn queues_per_thread(&self) -> Vec<u64> {
             vec![1, 1]
--- a/vhost/Cargo.toml
+++ b/vhost/Cargo.toml
@@ -23,6 +23,7 @@ vhost-net = ["vhost-kern"]
 vhost-user = []
 vhost-user-frontend = ["vhost-user"]
 vhost-user-backend = ["vhost-user"]
+gpu-socket = ["vhost-user"]
 xen = ["vm-memory/xen"]
 postcopy = []
 
--- a/vhost/src/vhost_user/backend_req_handler.rs
+++ b/vhost/src/vhost_user/backend_req_handler.rs
@@ -302,7 +302,7 @@ impl<T: VhostUserBackendReqHandlerMut> VhostUserBackendReqHandler for Mutex<T> {
 /// [BackendReqHandler]: struct.BackendReqHandler.html
 pub struct BackendReqHandler<S: VhostUserBackendReqHandler> {
     // underlying Unix domain socket for communication
-    main_sock: Endpoint<FrontendReq>,
+    main_sock: Endpoint<VhostUserMsgHeader<FrontendReq>>,
     // the vhost-user backend device object
     backend: Arc<S>,
 
@@ -319,7 +319,10 @@ pub struct BackendReqHandler<S: VhostUserBackendReqHandler> {
 
 impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
     /// Create a vhost-user backend endpoint.
-    pub(super) fn new(main_sock: Endpoint<FrontendReq>, backend: Arc<S>) -> Self {
+    pub(super) fn new(
+        main_sock: Endpoint<VhostUserMsgHeader<FrontendReq>>,
+        backend: Arc<S>,
+    ) -> Self {
         BackendReqHandler {
             main_sock,
             backend,
@@ -359,7 +362,10 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
     /// * - `path` - path of Unix domain socket listener to connect to
     /// * - `backend` - handler for requests from the frontend to the backend
     pub fn connect(path: &str, backend: Arc<S>) -> Result<Self> {
-        Ok(Self::new(Endpoint::<FrontendReq>::connect(path)?, backend))
+        Ok(Self::new(
+            Endpoint::<VhostUserMsgHeader<FrontendReq>>::connect(path)?,
+            backend,
+        ))
     }
 
     /// Mark endpoint as failed with specified error code.
@@ -965,7 +971,7 @@ mod tests {
     #[test]
     fn test_backend_req_handler_new() {
         let (p1, _p2) = UnixStream::pair().unwrap();
-        let endpoint = Endpoint::<FrontendReq>::from_stream(p1);
+        let endpoint = Endpoint::<VhostUserMsgHeader<FrontendReq>>::from_stream(p1);
         let backend = Arc::new(Mutex::new(DummyBackendReqHandler::new()));
         let mut handler = BackendReqHandler::new(endpoint, backend);
 
--- a/vhost/src/vhost_user/backend_req_handler.rs
+++ b/vhost/src/vhost_user/backend_req_handler.rs
@@ -12,6 +12,8 @@ use vm_memory::ByteValued;
 
 use super::backend_req::Backend;
 use super::connection::Endpoint;
+#[cfg(feature = "gpu-socket")]
+use super::gpu_backend_req::GpuBackend;
 use super::message::*;
 use super::{take_single_file, Error, Result};
 
@@ -65,6 +67,8 @@ pub trait VhostUserBackendReqHandler {
     fn get_config(&self, offset: u32, size: u32, flags: VhostUserConfigFlags) -> Result<Vec<u8>>;
     fn set_config(&self, offset: u32, buf: &[u8], flags: VhostUserConfigFlags) -> Result<()>;
     fn set_backend_req_fd(&self, _backend: Backend) {}
+    #[cfg(feature = "gpu-socket")]
+    fn set_gpu_socket(&self, gpu_backend: GpuBackend);
     fn get_inflight_fd(&self, inflight: &VhostUserInflight) -> Result<(VhostUserInflight, File)>;
     fn set_inflight_fd(&self, inflight: &VhostUserInflight, file: File) -> Result<()>;
     fn get_max_mem_slots(&self) -> Result<u64>;
@@ -124,6 +128,8 @@ pub trait VhostUserBackendReqHandlerMut {
     ) -> Result<Vec<u8>>;
     fn set_config(&mut self, offset: u32, buf: &[u8], flags: VhostUserConfigFlags) -> Result<()>;
     fn set_backend_req_fd(&mut self, _backend: Backend) {}
+    #[cfg(feature = "gpu-socket")]
+    fn set_gpu_socket(&mut self, _gpu_backend: GpuBackend);
     fn get_inflight_fd(
         &mut self,
         inflight: &VhostUserInflight,
@@ -233,6 +239,11 @@ impl<T: VhostUserBackendReqHandlerMut> VhostUserBackendReqHandler for Mutex<T> {
 
     fn set_backend_req_fd(&self, backend: Backend) {
         self.lock().unwrap().set_backend_req_fd(backend)
+    }
+
+    #[cfg(feature = "gpu-socket")]
+    fn set_gpu_socket(&self, gpu_backend: GpuBackend) {
+        self.lock().unwrap().set_gpu_socket(gpu_backend);
     }
 
     fn get_inflight_fd(&self, inflight: &VhostUserInflight) -> Result<(VhostUserInflight, File)> {
@@ -560,6 +571,11 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
                 let res = self.backend.set_inflight_fd(&msg, file);
                 self.send_ack_message(&hdr, res)?;
             }
+            #[cfg(feature = "gpu-socket")]
+            Ok(FrontendReq::GPU_SET_SOCKET) => {
+                let res = self.set_gpu_socket(files);
+                self.send_ack_message(&hdr, res)?;
+            }
             Ok(FrontendReq::GET_MAX_MEM_SLOTS) => {
                 self.check_proto_feature(VhostUserProtocolFeatures::CONFIGURE_MEM_SLOTS)?;
                 self.check_request_size(&hdr, size, 0)?;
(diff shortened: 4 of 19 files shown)


vhost/src/vhost_user/gpu_backend_req.rs
```diff
@@ -0,0 +1,54 @@
+use crate::vhost_user::connection::Endpoint;
```
stefano-garzarella: Copyright and SPDX missing

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -796,6 +802,17 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
...
     }
 
+    #[cfg(feature = "vhost-user-gpu-set-socket")]
+    fn set_gpu_socket(&mut self, files: Option<Vec<File>>) -> Result<()> {
+        let file = take_single_file(files).ok_or(Error::InvalidMessage)?;
+        // SAFETY: Safe because we have ownership of the files that were
+        // checked when received. We have to trust that they are Unix sockets
+        // since we have no way to check this. If not, it will fail later.
+        let _sock = unsafe { UnixStream::from_raw_fd(file.into_raw_fd()) };
+        // TODO: pass the socket to the backend
```
stefano-garzarella: I see this function modified later, can we add it directly in that commit?

vhost/src/vhost_user/gpu_backend_req.rs
```diff
@@ -1,13 +1,56 @@
-use crate::vhost_user::connection::Endpoint;
```
stefano-garzarella: Can we avoid those changes?

I mean following the https://rust-lang.github.io/rustfmt/?version=master&search=import#StdExternalCrate%5C%3A from the beginning

vhost/src/vhost_user/connection.rs
```diff
@@ -275,6 +275,42 @@ impl<H: MsgHeader> Endpoint<H> {
...
+    /// allows sending messages bigger than MAX_MSG_SIZE (up to MAX_OVERSIZED_MSG_SIZE)
+    /// # Return:
+    /// * - number of bytes sent on success
+    /// * - SocketRetry: temporary error caused by signals or short of resources.
+    /// * - SocketBroken: the underline socket is broken.
+    /// * - SocketError: other socket related errors.
+    /// * - OversizedMsg: message size is too big.
+    /// * - PartialMessage: received a partial message.
+    /// * - IncorrectFds: wrong number of attached fds.
+    pub fn send_message_oversized_with_payload<T: ByteValued>(
```
stefano-garzarella, line 314: Please split this commit in 2, one where you add send_message_oversized_with_payload() related stuff and one for update_scanout().

Also, if the oversized is supported only by gpu messages, can we move these new API there?

vhost/src/vhost_user/gpu_backend_req.rs
```diff
@@ -163,6 +163,18 @@ impl GpuBackend {
             .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("{}", e)))
     }
 
+    /// Send the VHOST_USER_GPU_GET_PROTOCOL_FEATURES message to the frontend and wait for a reply.
+    /// Get the supported protocol features bitmask.
+    pub fn get_protocol_features(&self) -> io::Result<VhostUserU64> {
+        self.send_header(GpuBackendReq::GET_PROTOCOL_FEATURES, None)
+    }
+
+    /// Send the VHOST_USER_GPU_GET_DISPLAY_INFO message to the frontend and wait for a reply.
```
stefano-garzarella, line 197: Wrong comment?

vhost/src/vhost_user/connection.rs
```diff
@@ -236,6 +236,31 @@ impl<H: MsgHeader> Endpoint<H> {
...
+    /// Send a message with header and body. Optional file descriptors may be attached to
+    /// the message.
+    ///
+    /// # Return:
+    /// * - number of bytes sent on success
+    /// * - SocketRetry: temporary error caused by signals or short of resources.
+    /// * - SocketBroken: the underline socket is broken.
+    /// * - SocketError: other socket related errors.
+    /// * - PartialMessage: received a partial message.
+    pub fn send_message_oversized<T: ByteValued>(
```
stefano-garzarella, line 248: Ditto, new API that are supposed to be used only for GPU protocol, should not be added and exposed with the vhost-user API. It's really confusing for future user why we have send_message and send_message_oversized.

Should be okay if we have a way to check if the device support it or not and return an error in that case.

stefano-garzarella: I also suggest adding the API for "oversized" message all together in a separate commit with a proper description in the commit message and also in the code comment.

vhost/src/vhost_user/gpu_message.rs
```diff
@@ -342,7 +342,7 @@ unsafe impl ByteValued for VhostUserGpuDMABUFScanout2 {}
 impl VhostUserMsgValidator for VhostUserGpuDMABUFScanout2 {}
 
 /// The VhostUserGpuCursorPos from the QEMU vhost-user-gpu specification.
-#[derive(Copy, Clone, Debug, Default)]
+#[derive(Copy, Clone, Debug, Default, PartialEq)]
```
stefano-garzarella, line 345: Add this when introducing `VhostUserGpuCursorPos` and `VhostUserGpuCursorUpdate`

stefano-garzarella (commented):
I left some comments, but the code is almost fine with me.

Please fix the commit messages, some of them doesn't have any description (please also just a line describing the type of message you're adding is enough, but avoid empty), others doesn't have new lines before S-o-b.

Also check copyright and SPDX of new files.

vhost/src/vhost_user/connection.rs
dorindabassey: yeah, I agree. Since this methods are specific for the GPU protocol, maybe we can put it behind a gpu feature.

vhost-user-backend/Cargo.toml
```diff
@@ -11,6 +11,7 @@ license = "Apache-2.0"
 [features]
 xen = ["vm-memory/xen", "vhost/xen"]
 postcopy = ["vhost/postcopy", "userfaultfd"]
+gpu-set-socket = ["vhost/vhost-user-gpu-set-socket"]
```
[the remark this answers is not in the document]
dorindabassey: this feature was added for the `gpu-set-socket` functionality. Given that in QEMU for vhost-user-gpu the expectation is different compared to other backend, it requires a VHOST_USER_GPU_SET_SOCKET to obtain the socket. Also this helps to manage the dependencies for the vost-device-gpu crate, since other crates don't need it.

vhost/src/vhost_user/connection.rs
stefano-garzarella: I'm not sure a feature is enough. What I mean is that if these functions are not meant to be used by any device, then I would avoid putting in the common modules, but in a gpu-specific module, otherwise device implementers can be confused and use this API.

Otherwise we have to at least put a check of the negotiated virtio features and return an error if this API is not supported.

vhost/src/vhost_user/gpu_backend_req.rs
dorindabassey: It seems you are referencing old commit here, but IIUC you mean grouping related imports together?

vhost-user-backend/Cargo.toml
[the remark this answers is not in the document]
stefano-garzarella: What dependencies depend on this feature?

If we save little code and no dependencies, I don't know how worthwhile it is.

vhost/src/vhost_user/connection.rs
mtjhrc: `Endpoint` is internal to this crate, it is not a public API, so this shouldn't really be a problem.

As far as I can tell there is not really a message size limit in the protocol itself, this is just an internal limitation of `Endpoint` implementation, see this comment:
https://github.com/rust-vmm/vhost/blob/bd6bf1325b4930d4123f6154ce755bd5289c617b/vhost/src/vhost_user/message.rs#L26-L33

Ideally I think it could be nice to not have this limit and handle arbitrary sized messages, but I didn't want to rewrite the whole `Endpoint` just for this.  If the other side (a frontend) of the  vhost user gpu socket protocol is implemented in this crate, this would likely be necessary. Since this is an internal API, I hope this would be fine for now.

stefano-garzarella: okay, I see.

I don't remember why we have that limit on the sender side. On the receiver side it looks like it's a limit to avoid pre-allocation, but IMHO we can remove that limit in the sender. If the receiver doesn't like the size, can always send back an error.

I mean, we can just check if the len is on 32bit (as you did), without adding new API. WDYT?

vhost-user-backend/Cargo.toml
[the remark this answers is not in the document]
mtjhrc: The `gpu-set-socket` is here to enable the  `vhost-user-gpu-set-socket` feature in the `vhost` dependency. There aren't any external dependencies this brings in. I put it behind a feature flag, because this feature is useless for every other device apart from a gpu device.

I can remove this flag and have it be always enabled if you want. If we decide to keep it I'll make sure to add it to CI.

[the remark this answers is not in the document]
stefano-garzarella: > The `gpu-set-socket` is here to enable the `vhost-user-gpu-set-socket` feature in the `vhost` dependency. There aren't any external dependencies this brings in. I put it behind a feature flag, because this feature is useless for every other device apart from a gpu device.

Yep, I meant both features. Sorry to be unclear.

> 
> I can remove this flag and have it be always enabled if you want. If we decide to keep it I'll make sure to add it to CI.

I'm in doubt, on the one hand I understand why, on the other hand we don't want to add too many features. But maybe in this case it makes sense, because a generic backend doesn't want to handle those types of messages at all.

If we want to keep it, what about a shorter name like `gpu` ?

vhost/src/vhost_user/connection.rs
mtjhrc: Yes sure, I can rewrite it like that. I maybe see one reason it is the way it is: for testing the backend against fronted in this crate, this might have been like an internal assertion when writing tests (testing that the test itself is OK) But this isn't really that useful, since the receiving side would hopefully fail anyway.

vhost-user-backend/Cargo.toml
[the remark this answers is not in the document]
mtjhrc: Sure, how about naming the feature `gpu-socket` in both `vhost` and `vhost-user-backend` ? This seem a bit more clear than `gpu` alone.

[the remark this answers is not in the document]
stefano-garzarella: > Sure, how about naming the feature `gpu-socket` in both `vhost` and `vhost-user-backend` ? This seem a bit more clear than `gpu` alone.

yes, I was going to write "or gpu-socket" in the previous message, so if you also think that is okay, I agree to use `gpu-socket` for both ;-)

mtjhrc:
Apart from changes you requested I also changed the signature of cursor_update, to accept the array outside of the struct as I realized this is more practical to the user. This also meant I removed a test for sending such a big message body.

I am not sure about the changes in regards to sending big message. I added a constant to the `MsgHeader` trait and use it for checking size when sending messages.  I also left a TODO comment. I think we could support u32::MAX everywhere but, this warrants a separate PR and a discussion if we want to enforce the current MAX_MSG_SIZE for `vhost` protocol, also there are some redundant checks for this at multiple levels and I think the checks could probably overflow when adding numbers.

mtjhrc:
I am not sure about the changes in regards to sending big message. I added a constant to the `MsgHeader` trait and use it for checking size when sending messages.  I also left a TODO comment. I think we could support u32::MAX everywhere but, this warrants a separate PR and a discussion if we want to enforce the current MAX_MSG_SIZE for `vhost` protocol, also there are some redundant checks for this at multiple levels and I think the checks could probably overflow when adding numbers.

vhost-user-backend/src/backend.rs
```diff
@@ -84,6 +87,9 @@ pub trait VhostUserBackend: Send + Sync {
     /// function.
     fn set_backend_req_fd(&self, _backend: Backend) {}
 
+    #[cfg(feature = "gpu-socket")]
+    fn set_gpu_socket(&self, _gpu_backend: GpuBackend) {}
```
stefano-garzarella, line 91: What about adding a documentation here, like for other methods?

vhost/Cargo.toml
```diff
@@ -23,7 +23,7 @@ vhost-net = ["vhost-kern"]
 vhost-user = []
 vhost-user-frontend = ["vhost-user"]
 vhost-user-backend = ["vhost-user"]
-gpu-socket = ["vhost-user"]
+gpu-socket = ["vhost-user", "vhost-user-backend"]
```
stefano-garzarella, line 26: Why do we need this dependency?
For example could we allow to use this crate only for frontend, but with gpu support ?

vhost/src/vhost_user/gpu_backend_req.rs
```diff
@@ -41,6 +41,20 @@ impl BackendInternal {
         self.wait_for_ack(&hdr)
     }
 
+    fn send_message_no_reply<T: ByteValued>(
```
stefano-garzarella, line 44: Why not just adding a parameter to `send_message()`.

That said, if `VHOST_USER_PROTOCOL_F_REPLY_ACK` is negotiated, usually all the messages should expect an ACK, is this an exception in vhost-user-gpu? Maybe we should document this.

vhost/src/vhost_user/gpu_backend_req.rs
```diff
@@ -1,17 +1,59 @@
...
+        self.sock.send_header(&hdr, fds)?;
+
+        self.wait_for_ack(&hdr)
+    }
+
+    fn wait_for_ack<V: ByteValued + Sized + Default + VhostUserMsgValidator>(
+        &mut self,
+        hdr: &VhostUserGpuMsgHeader<GpuBackendReq>,
+    ) -> vhost_user::Result<V> {
+        self.check_state()?;
```
stefano-garzarella, line 92: Should we check if VHOST_USER_PROTOCOL_F_REPLY_ACK is negotiated?

Or the ack is always expected?

stefano-garzarella (commented):
I did a quick review and I think we are in a good shape :-)

I just suggest that you fix the description a bit in the commits. Take a look here: https://docs.kernel.org/process/submitting-patches.html

> Describe your changes in imperative mood, e.g. “make xyzzy do frotz” instead of “[This patch] makes xyzzy do frotz” or “[I] changed xyzzy to do frotz”, as if you are giving orders to the codebase to change its behaviour.

aesteve-rh (changes requested):
In general LGTM, but the PR is really long, not sure if it would be better to split some of the initial commits that have fewer dependencies.

I left a few comments, and I will give it another pass for the next round.

vhost/src/vhost_user/gpu_backend_req.rs
mtjhrc: There is no `VHOST_USER_PROTOCOL_F_REPLY_ACK` flag in this protocol, so we don't expect acks for every message. 
Having a parameter for ack, would be messy I think, because this method has an generic argument for the type of reply, and the type of reply has to logically be `()` when not waiting for reply. Instead I think it could be achieved by having a special return type for no-ack messages with some trait. 

Another alternative could be calling `wait_for_ack` in the the public command methods depending if the command waits or doesn't wait.

vhost/src/vhost_user/gpu_backend_req.rs
mtjhrc: The ack is always expected for some messages and never expected for other messages. There is no `VHOST_USER_PROTOCOL_F_REPLY_ACK` flag for this protocol. I'll make sure to add a comment in the source code.

vhost/src/vhost_user/gpu_backend_req.rs
stefano-garzarella: > Another alternative could be calling wait_for_ack in the the public command methods depending if the command waits or doesn't wait.

Yeah, I like this. 
So IIUC, some messages need a reply, so maybe instead of `wait_for_ack` we can call it `recv_reply`.

So the API will be:
- `send_header()`
- `send_message()`
- `send_message_with_payload()`
- `recv_reply()`

And call `recv_reply()` where needed. (I'd remove it also from `send_header()`).

IMHO this makes the code a little more understandable and reviewable, going to check for each message when a reply is required or not.

mtjhrc: After changing it to this. Do you think it is still enough to have tests just for this internal API (`send_header()`, `send_message()`, `send_message_with_payload()`, `recv_reply()`) ? Or should I change the tests to test the all of the public methods instead of internal API?

stefano-garzarella: > Hmm, actually I just realized there could be a theoretical issue with this. Because unlocking the mutex between the calls to `send_message()` and `recv_reply()` could theoretically not be correct, depending on the frontend implementation. Because in between these calls another `send_message()` could be issued and we could receive a reply for that one. As far as I can tell, this cannot happen in current QEMU implementation though, because messages are not processed in parallel. If we don't care about this I wonder if we even need to wrap this in a mutex.

Does the specification cover this case?

I thought in general we don't support sending two requests without waiting for a response to the first one first, in any case I think your approach is good.


> Doing this with the mutex would mean doing something like this which I guess is not that bad:

Yep, I think this is fine.

vhost/src/vhost_user/gpu_message.rs
```diff
@@ -0,0 +1,177 @@
...
+            .field("size", &{ self.size })
+            .finish()
+    }
+}
+
+impl<R: Req> Clone for VhostUserGpuMsgHeader<R> {
+    fn clone(&self) -> VhostUserGpuMsgHeader<R> {
+        *self
+    }
+}
```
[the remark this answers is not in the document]
mtjhrc: Fixed for `VhostUserGpuMsgHeader` - yes, this was pointless. I just copied it from VhostUserMsgHeader, which has the same problem.

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -65,6 +67,8 @@ pub trait VhostUserBackendReqHandler {
     fn get_config(&self, offset: u32, size: u32, flags: VhostUserConfigFlags) -> Result<Vec<u8>>;
     fn set_config(&self, offset: u32, buf: &[u8], flags: VhostUserConfigFlags) -> Result<()>;
     fn set_backend_req_fd(&self, _backend: Backend) {}
+    #[cfg(feature = "gpu-socket")]
+    fn set_gpu_socket(&self, _gpu_backend: GpuBackend) {}
```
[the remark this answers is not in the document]
mtjhrc: Yes, I agree, this doesn't seem necessary since this an optional feature.

vhost/Cargo.toml
mtjhrc: Yes I think you are right, we don't really need this dependency.

mtjhrc:
I changed the internal API used here like discussed here with @stefano-garzarella, this makes each message method explicitly call `recv_reply` (instead of `wait_for_ack`) if it is needed. 
So instead of 
```Rust
pub fn get_display_info(&self) -> io::Result<VirtioGpuRespDisplayInfo> {
    self.send_header(GpuBackendReq::GET_DISPLAY_INFO, None)
}
```
the public methods now do:
```Rust
pub fn get_display_info(&self) -> io::Result<VirtioGpuRespDisplayInfo> {
    let node = self.node(); // locks the mutex and returns a reference 
   
    let hdr = node.send_header(GpuBackendReq::GET_DISPLAY_INFO, None)?;
    node.recv_reply(&hdr)
}
```
Note that some messeges aren't supposed to wait for a reply so this removes the need for overly specific utility methods like `send_message_with_payload_no_reply`.
I also added a comment in the code explaining that there is no  `VHOST_USER_PROTOCOL_F_REPLY_ACK` flag for this protocol. Because of this I now changed the tests to call the public methods, since more of the code moved there.

stefano-garzarella:
@mtjhrc great! Thanks, I'll review in the next days (I'm going to DevConf at Brno, so maybe it will take a while, sorry in advance).

Since you are here, can you also rebase on the latest main?

mtjhrc:
> @mtjhrc great! Thanks, I'll review in the next days (I'm going to DevConf at Brno, so maybe it will take a while, sorry in advance).
> 
> Since you are here, can you also rebase on the latest main?

Rebased. No problem, also sorry for the delay fixing this on my side too.

stefano-garzarella (commented):
@mtjhrc great work about send_*() API!

The PR LGTM, but maybe during the rebase the change to enable `gpu-socket` in `.buildkite/rust-vmm-ci-tests.json ` are lost. Can you check it?

aesteve-rh (approved):
LGTM!

mtjhrc:
> @mtjhrc great work about send_*() API!
> 
> The PR LGTM, but maybe during the rebase the change to enable `gpu-socket` in `.buildkite/rust-vmm-ci-tests.json ` are lost. Can you check it?

I have fixed this and rebased the PR.

stefano-garzarella (approved):

stefano-garzarella:
@germag @slp can you take a look?

slp (approved):
LGTM, great work here.

PR #240, opened 2024-04-24
vhost-user-backend: set event_idx boolean at set_features

eugpermar:
A frontend can skip the call to set_vring_base, assuming the ring will be initialized at 0.  If that frontend acknowledge EVENT_IDX vring and VhostUserHandler backend will never know they need to use event idx.

Move the features configuration of the vring and the backend to set_features method.

---
I'm not 100% sure we should remove from set_vring_base.  QEMU history shows SET_FEATURES and SET_VRING_BASE were added by the same patch, so I guess there are no frontends not sending SET_FEATURES.  Even if it is using a transitional device.

At least, it is way more unlikely than frontends skipping SET_VRING_BASE.

Not adding unit/integration test. A guide on how to do that would be useful. Due to that, marking as a draft.

### DIFF
--- a/vhost-user-backend/src/handler.rs
+++ b/vhost-user-backend/src/handler.rs
@@ -279,6 +279,11 @@ where
             }
         }
 
+        let event_idx: bool = (self.acked_features & (1 << VIRTIO_RING_F_EVENT_IDX)) != 0;
+        for vring in self.vrings.iter_mut() {
+            vring.set_queue_event_idx(event_idx);
+        }
+        self.backend.set_event_idx(event_idx);
         self.backend.acked_features(self.acked_features);
 
         Ok(())
@@ -398,11 +403,7 @@ where
             .get(index as usize)
             .ok_or_else(|| VhostUserError::InvalidParam)?;
 
-        let event_idx: bool = (self.acked_features & (1 << VIRTIO_RING_F_EVENT_IDX)) != 0;
-
         vring.set_queue_next_avail(base as u16);
-        vring.set_queue_event_idx(event_idx);
-        self.backend.set_event_idx(event_idx);
 
         Ok(())
     }
--- a/vhost-user-backend/CHANGELOG.md
+++ b/vhost-user-backend/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Added
 
 ### Changed
+- [[#240]](https://github.com/rust-vmm/vhost/pull/240) Move the set of event_idx property from set_vring_base callback to set_features one
 
 ### Fixed
 


germag (approved):

vhost-user-backend/src/handler.rs
```diff
@@ -279,6 +279,11 @@ where
             }
         }
 
+        let event_idx: bool = (self.acked_features & (1 << VIRTIO_RING_F_EVENT_IDX)) != 0;
+        for vring in self.vrings.iter_mut() {
+            vring.set_queue_event_idx(event_idx);
+        }
+        self.backend.set_event_idx(event_idx);
```
stefano-garzarella, line 286: Should we do this after setting the acked features in the backend?

I checked some backend and no one checks negotiated features (I don't know if we should fix this), but in theory it could avoid enabling event_idx if the feature is not negotiated.

stefano-garzarella (commented):
> I'm not 100% sure we should remove from set_vring_base.  QEMU history
shows SET_FEATURES and SET_VRING_BASE were added by the same patch, so I
guess there are no frontends not sending SET_FEATURES.  Even if it is
using a transitional device.
At least, it is way more unlikely than frontends skipping SET_VRING_BASE.

Yeah, I think this is fine.

BTW github doesn't remove stuff after `---`, so I suggest you editing the commit message before merging it if you agree.

vhost-user-backend/src/handler.rs
eugpermar: To be honest I'm not sure why we need an independent call (plus vring ones) instead of the backend checking for _F_EVENT_IDX set. What you say makes sense to me, but I can also see how a backend could simply check for acked features and then do not allow to modify it afterwards.

So I'm happy to other them the way is more convenient for sure. Should I reverse them?

Thanks!

stefano-garzarella: > To be honest I'm not sure why we need an independent call (plus vring ones) instead of the backend checking for _F_EVENT_IDX set. What you say makes sense to me, but I can also see how a backend could simply check for acked features and then do not allow to modify it afterwards.

I've been asking myself the same question, and I don't have an answer. :-) It would be nice to fix it, but not in this PR obviously.

> 
> So I'm happy to other them the way is more convenient for sure. Should I reverse them?

As you prefer, I also don't have a strong opinion. I think the best way is to avoid the independent call as you mention, but we can do later.

Thanks!

germag: Agreed, IMO the `backend.set_event_idx(event_idx)`API is redundant. And I prefer to be called before `        self.backend.acked_features(self.acked_features);`, but it doesn't really matter, the backend can always implement a nop `set_event_idx()`. I also this we should remove it in the future

stefano-garzarella: Okay, let's merge this and we can fix it in the future.

stefano-garzarella:
I just removed the part after `---`

stefano-garzarella (approved):

PR #241, opened 2024-05-21
Add shared object

aesteve-rh:
### Summary of the PR

Add `SHARED_OBJECT_*` vhost-user message IDs to align with standard at
https://qemu-project.gitlab.io/qemu/interop/vhost-user.html

Move `FS_*` type IDs to avoid current and future collisions.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost/src/vhost_user/backend_req.rs
+++ b/vhost/src/vhost_user/backend_req.rs
@@ -19,6 +19,9 @@ struct BackendInternal {
     // Protocol feature VHOST_USER_PROTOCOL_F_REPLY_ACK has been negotiated.
     reply_ack_negotiated: bool,
 
+    // Protocol feature VHOST_USER_PROTOCOL_F_SHARED_OBJECT has been negotiated.
+    shared_object_negotiated: bool,
+
     // whether the endpoint has encountered any failure
     error: Option<i32>,
 }
@@ -88,6 +91,7 @@ impl Backend {
             node: Arc::new(Mutex::new(BackendInternal {
                 sock: ep,
                 reply_ack_negotiated: false,
+                shared_object_negotiated: false,
                 error: None,
             })),
         }
@@ -122,6 +126,14 @@ impl Backend {
         self.node().reply_ack_negotiated = enable;
     }
 
+    /// Set the negotiation state of the `VHOST_USER_PROTOCOL_F_SHARED_OBJECT` protocol feature.
+    ///
+    /// When the `VHOST_USER_PROTOCOL_F_SHARED_OBJECT` protocol feature has been negotiated,
+    /// the backend is allowed to send "SHARED_OBJECT_*" messages to the frontend.
+    pub fn set_shared_object_flag(&self, enable: bool) {
+        self.node().shared_object_negotiated = enable;
+    }
+
     /// Mark endpoint as failed with specified error code.
     pub fn set_failed(&self, error: i32) {
         self.node().error = Some(error);
@@ -131,11 +143,23 @@ impl Backend {
 impl VhostUserFrontendReqHandler for Backend {
     /// Forward vhost-user shared-object add request to the frontend.
     fn shared_object_add(&self, uuid: &VhostUserSharedMsg) -> HandlerResult<u64> {
+        if !self.node().shared_object_negotiated {
+            return Err(io::Error::new(
+                io::ErrorKind::Other,
+                "Shared Object feature not negotiated",
+            ));
+        }
         self.send_message(BackendReq::SHARED_OBJECT_ADD, uuid, None)
     }
 
     /// Forward vhost-user shared-object remove request to the frontend.
     fn shared_object_remove(&self, uuid: &VhostUserSharedMsg) -> HandlerResult<u64> {
+        if !self.node().shared_object_negotiated {
+            return Err(io::Error::new(
+                io::ErrorKind::Other,
+                "Shared Object feature not negotiated",
+            ));
+        }
         self.send_message(BackendReq::SHARED_OBJECT_REMOVE, uuid, None)
     }
 
@@ -145,6 +169,12 @@ impl VhostUserFrontendReqHandler for Backend {
         uuid: &VhostUserSharedMsg,
         fd: &dyn AsRawFd,
     ) -> HandlerResult<u64> {
+        if !self.node().shared_object_negotiated {
+            return Err(io::Error::new(
+                io::ErrorKind::Other,
+                "Shared Object feature not negotiated",
+            ));
+        }
         self.send_message(
             BackendReq::SHARED_OBJECT_LOOKUP,
             uuid,
--- a/vhost/src/vhost_user/backend_req.rs
+++ b/vhost/src/vhost_user/backend_req.rs
@@ -129,6 +129,29 @@ impl Backend {
 }
 
 impl VhostUserFrontendReqHandler for Backend {
+    /// Forward vhost-user shared-object add request to the frontend.
+    fn shared_object_add(&self, uuid: &VhostUserSharedMsg) -> HandlerResult<u64> {
+        self.send_message(BackendReq::SHARED_OBJECT_ADD, uuid, None)
+    }
+
+    /// Forward vhost-user shared-object remove request to the frontend.
+    fn shared_object_remove(&self, uuid: &VhostUserSharedMsg) -> HandlerResult<u64> {
+        self.send_message(BackendReq::SHARED_OBJECT_REMOVE, uuid, None)
+    }
+
+    /// Forward vhost-user shared-object lookup request to the frontend.
+    fn shared_object_lookup(
+        &self,
+        uuid: &VhostUserSharedMsg,
+        fd: &dyn AsRawFd,
+    ) -> HandlerResult<u64> {
+        self.send_message(
+            BackendReq::SHARED_OBJECT_LOOKUP,
+            uuid,
+            Some(&[fd.as_raw_fd()]),
+        )
+    }
+
     /// Forward vhost-user-fs map file requests to the backend.
     fn fs_backend_map(&self, fs: &VhostUserFSBackendMsg, fd: &dyn AsRawFd) -> HandlerResult<u64> {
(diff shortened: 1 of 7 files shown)


vhost/src/vhost_user/message.rs
```diff
@@ -203,14 +203,16 @@ enum_value! {
...
+        FS_MAP = 1000,
         /// Virtio-fs draft: unmap file content from the window.
-        FS_UNMAP = 7,
+        FS_UNMAP = 1001,
         /// Virtio-fs draft: sync file content.
-        FS_SYNC = 8,
+        FS_SYNC = 1002,
         /// Virtio-fs draft: perform a read/write from an fd directly to GPA.
-        FS_IO = 9,
+        FS_IO = 1003,
```
stefano-garzarella, line 223: @germag can you confirm this is fine for virtio-fs?

(not for this PR) Maybe we should add a feature to enable/disable draft code.

vhost/src/vhost_user/message.rs
```diff
@@ -948,6 +956,25 @@ enum_value! {
...
 
+/// Contains UUID to interact with associated virtio shared object.
+#[repr(C)]
+#[derive(Clone, Copy, Default)]
+pub struct VhostUserSharedMsg {
+    /// UUID of the shared object
+    pub uuid: Uuid,
+}
+
+// SAFETY: Safe because VhostUserTransferDeviceState is a POD
```
stefano-garzarella, line 967: `VhostUserTransferDeviceState` should be replaced

stefano-garzarella: Should we also mention that `Uuid` is a POD?

Following the definitions:
```
pub struct Uuid(Bytes);

pub type Bytes = [u8; 16];
```

vhost/src/vhost_user/frontend_req_handler.rs
```diff
@@ -428,6 +535,12 @@ mod tests {
             let res = handler.handle_request().unwrap();
             assert_eq!(res, 0);
             handler.handle_request().unwrap_err();
+            assert_eq!(handler.handle_request().unwrap(), 0);
+            assert_eq!(handler.handle_request().unwrap(), 1);
+            assert_eq!(handler.handle_request().unwrap(), 0);
+            assert_eq!(handler.handle_request().unwrap(), 1);
+            assert_eq!(handler.handle_request().unwrap(), 0);
+            assert_eq!(handler.handle_request().unwrap(), 1);
```
stefano-garzarella, line 543: What about adding a comment that these tests are for shared objects?

BTW I tried to change the value asserted and the test passes in any case, so I think we should `join()` it in some way to check properly. (not for this PR)

vhost/src/vhost_user/message.rs
```diff
@@ -203,6 +205,12 @@ enum_value! {
         VRING_CALL = 4,
         /// Indicate that an error occurred on the specific vring.
         VRING_ERR = 5,
+        /// Add a virtio shared object.
+        SHARED_OBJECT_ADD = 6,
+        /// Remove a virtio shared object.
+        SHARED_OBJECT_REMOVE = 7,
+        /// Lookup for a virtio shared object.
+        SHARED_OBJECT_LOOKUP = 8,
```
stefano-garzarella, line 213: Should we also add `SHARED_OBJECT` in `VhostUserProtocolFeatures`?

And maybe check it while sending/receiving these messages.

vhost/src/vhost_user/message.rs
aesteve-rh: Yeah, sorry. Copy/pasted and then forgot to review the comment before posting.

I guess we can say it is safe because it is a fixed size array?

vhost/src/vhost_user/frontend_req_handler.rs
aesteve-rh: I'll add the comment.

I would've sworn I saw assertion errors here. I'll try again.

Also, I can open a follow-up PR to improve the logic of this test (if I do not improve it for this PR already).

vhost/src/vhost_user/message.rs
stefano-garzarella: > Yeah, sorry. Copy/pasted and then forgot to review the comment before posting.

Don't worry ;-)
 
> I guess we can say it is safe because it is a fixed size array?

Yep, I think so.

vhost/src/vhost_user/frontend_req_handler.rs
stefano-garzarella: > I'll add the comment.
> 
> I would've sworn I saw assertion errors here. I'll try again.

It could be that I did something wrong while testing. ;-)

> 
> Also, I can open a follow-up PR to improve the logic of this test (if I do not improve it for this PR already).

That would be great! Thanks!

aesteve-rh: > It could be that I did something wrong while testing. ;-)

You did not, it indeed requires joining. Which makes sense. But I checked and the first `asser_eq` that was there for the `FS_*` messages catches the fail without joining, so that is what threw me off. I failed to understand the difference... But I can push the `join()` already for this commit.

vhost/src/vhost_user/message.rs
alyssais: They were only used for DAX, right?

stefano-garzarella: @germag yeah, now I remember, so it should be fine for this PR, but we should remove them or use a no-default feature to hide "draft" code.

@alyssais yep, I think so.

vhost/src/vhost_user/backend_req.rs
```diff
@@ -131,11 +143,17 @@ impl Backend {
 impl VhostUserFrontendReqHandler for Backend {
     /// Forward vhost-user shared-object add request to the frontend.
     fn shared_object_add(&self, uuid: &VhostUserSharedMsg) -> HandlerResult<u64> {
+        if !self.node().shared_object_negotiated {
+            return Ok(1);
```
stefano-garzarella, line 147: Should we return an error?

aesteve-rh: I hesitated here.

So the spec states: `the front-end must respond with zero when operation is successfully completed, or non-zero otherwise.`, so returning `Ok(1)` is interpreted as an error by the device.

On the other hand, this is not the frontend setting the return value. So maybe it would be better to return an error. But if I do, which one? Should I create one `ErrorType`?

stefano-garzarella: IIUC these are the APIs called from the backend, so before we even send the message to the frontend. So I think it's an implementation detail of ours.

What about the `Error::InactiveFeature(VhostUserVirtioFeatures)` already defined in https://github.com/rust-vmm/vhost/blob/043d0d8fa2037da47f4a621da6c13d5f379b7c69/vhost/src/vhost_user/mod.rs#L64

aesteve-rh: Yes, this occurs before sending the message to the frontend.

Uhm, it needs to be an IoError... What about `return Err(std::io::Error::from(ErrorKind::InvalidInput));`?

stefano-garzarella: Ah, right, or `ErrorKind::Unsupported`. Is up to you ;-)

aesteve-rh: Pushed with `ErrorKind::Other` in the end. `ErrorKind::Unsupported` would be a possibility but sounds like it is not supported at all to send those messages, so I thought that better to be explicit and let the device know the exact reason.

vhost/src/vhost_user/frontend_req_handler.rs
stefano-garzarella: @aesteve-rh what about this? Will we fix in another PR?

aesteve-rh: Sorry, I'm not following. I've added the `join()` and a comment before the assertions to separate them from the other. Now if the check fails the test will surely fail. What is there left to fix?

In any case, whatever is left here to do related with the shared object feature, I'd say better to fix it here. Otherwise, e.g., improving test infra in general, it can go in another PR.

stefano-garzarella: Ooops, I missed that. I suggest adding a changelog on every update, like for mail-based workflow.

Also, since it fixes a pre-existing issue in our tests, what about having a separate commit with an explanation?

aesteve-rh:
v3 -> v4: (I think) Separate the `frontend_req_handler` test fix (joining the frontend handler threads) to its own commit.

stefano-garzarella:
Thanks, LGTM!

Last doubt and whether we should merge commit 3 and 4 together to improve bisectability.

aesteve-rh:
> Last doubt: whether we should merge commit 3 and 4 together to improve bisectability.

I do not have a strong opinion either way. Commit 4 updated `BackendInternal` and `vhost-user-backend`, and did not change any parts of the shared object logic, except for the check itself. So it felt a contained feature that would be clearer to review in a separate commit. On the other hand, it is still part of the shared object support, so I see your point.

If in doubt, I can squash them as they are already reviewed anyway.

stefano-garzarella:
> > Last doubt: whether we should merge commit 3 and 4 together to improve bisectability.
> 
> I do not have a strong opinion either way. Commit 4 updated `BackendInternal` and `vhost-user-backend`, and did not change any parts of the shared object logic, except for the check itself. So it felt a contained feature that would be clearer to review in a separate commit. On the other hand, it is still part of the shared object support, so I see your point.
> 
> If in doubt, I can squash them as they are already reviewed anyway.

I don't have a strong opinion either. Theoretically, patch 3 without patch 4 is incorrect, but only if the backend misbehaves. Also, the frontend should handle this eventually, so I think it's fine unless others have different opinions.

stefano-garzarella (approved):

stefano-garzarella:
My approval is still up :-) any other reviews? @Ablu @eryugey @germag @jiangliu @sboeuf @slp

slp (approved):
Confirmed it matches QEMU implementation. LGTM.

PR #243, opened 2024-06-08
vhost_user: ignore unknown features from backend

alyssais:
### Summary of the PR

virtiofsd 1.11.0 added support for VHOST_USER_PROTOCOL_F_DEVICE_STATE. Upgrading virtiofsd meant that the latest released version of Cloud Hypervisor (39.0) was no longer able to communicate with it, because rather than just ignoring the unsupported feature, it got an unrecoverable "invalid message" error.  This demonstrates that it's better for frontends to just ignore offers of unsupported features. If the backend requires some feature, it'll get a chance to know that when we send VHOST_USER_SET_PROTOCOL_FEATURES anyway.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost/src/vhost_user/frontend.rs
+++ b/vhost/src/vhost_user/frontend.rs
@@ -362,12 +362,9 @@ impl VhostUserFrontend for Frontend {
         let hdr = node.send_request_header(FrontendReq::GET_PROTOCOL_FEATURES, None)?;
         let val = node.recv_reply::<VhostUserU64>(&hdr)?;
         node.protocol_features = val.value;
-        // Should we support forward compatibility?
-        // If so just mask out unrecognized flags instead of return errors.
-        match VhostUserProtocolFeatures::from_bits(node.protocol_features) {
-            Some(val) => Ok(val),
-            None => error_code(VhostUserError::InvalidMessage),
-        }
+        Ok(VhostUserProtocolFeatures::from_bits_truncate(
+            node.protocol_features,
+        ))
     }
 
     fn set_protocol_features(&mut self, features: VhostUserProtocolFeatures) -> Result<()> {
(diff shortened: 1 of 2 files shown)


stefano-garzarella (dismissed):
Yeah, thanks for this fix, I completely agree on this change!

vhost/src/vhost_user/frontend.rs
```diff
@@ -935,7 +932,7 @@ mod tests {
 
         let pfeatures = VhostUserProtocolFeatures::all();
         let hdr = VhostUserMsgHeader::new(FrontendReq::GET_PROTOCOL_FEATURES, 0x4, 8);
-        let msg = VhostUserU64::new(pfeatures.bits());
+        let msg = VhostUserU64::new(pfeatures.bits() | 1 << 63);
```
aesteve-rh, line 935: nit: Maybe worth adding a comment to mention that we test that only supported features are returned (iiuc it is the reason for setting this bit) or something along those lines?

Other than that, lgtm :)

stefano-garzarella (dismissed):

aesteve-rh (approved):
Thanks!

vhost/src/vhost_user/frontend.rs
```diff
@@ -935,7 +932,8 @@ mod tests {
 
         let pfeatures = VhostUserProtocolFeatures::all();
         let hdr = VhostUserMsgHeader::new(FrontendReq::GET_PROTOCOL_FEATURES, 0x4, 8);
-        let msg = VhostUserU64::new(pfeatures.bits());
+        // Unknown feature bits should be ignored.
+        let msg = VhostUserU64::new(pfeatures.bits() | 1 << 63);
```
germag, line 936: I'm not sure the comment is enough, can a const be created?, with some descriptive name like `const FAKE_FEATURE: u64 = 1 << 63`  instead, because it could be interpreted that setting that bit, enables a feature to ignore features.

besides that LGTM

alyssais: Done.

stefano-garzarella (approved):

germag (approved):

Issue #213, opened 2023-11-20
Remove non-standard BackedReq types


germag:
The following back-end request types are not part of the vhost-user protocol and should be removed:

```rust
pub enum BackendReq {
    ...
    /// Virtio-fs draft: map file content into the window.
    FS_MAP = 6,
    /// Virtio-fs draft: unmap file content from the window.
    FS_UNMAP = 7,
    /// Virtio-fs draft: sync file content.
    FS_SYNC = 8,
    /// Virtio-fs draft: perform a read/write from an fd directly to GPA.
    FS_IO = 9,
    ...
}
```

These are not in the vhost-user standard, which in itself is confusing, but even conflict with values that are defined in the standard (see [Back-end message types](https://qemu-project.gitlab.io/qemu/interop/vhost-user.html#id37)), such as: VHOST_USER_BACKEND_SHARED_OBJECT_ADD, VHOST_USER_BACKEND_SHARED_OBJECT_REMOVE and VHOST_USER_BACKEND_SHARED_OBJECT_LOOKUP.

This issue is to continue the discussion started on #205

aesteve-rh:
This can be closed with https://github.com/rust-vmm/vhost/pull/241

stefano-garzarella:
@aesteve-rh I don't know if we can do even more by removing those types completely, or putting them behind a feature (e.g. unstable)

aesteve-rh:
Ah, fair point. It is true that the issue states removing them, not only "moving them out of the way".

So I'd be fine either way, but I see some value in having some support for non-standard features (as long as they are backed up by a patch or an effort to standardize, or something like that). So I guess I'd vote for the feature option.

Given that we have `staging` devices in `vhost-device`, and these may require non-standard messages, having these devices upstream would be supported just by enabling the feature in the crate settings.

Not sure if it makes much sense to keep the `FS_*` messages specifically, though...

stefano-garzarella:
> Given that we have `staging` devices in `vhost-device`, and these may require non-standard messages, having these devices upstream would be supported just by enabling the feature in the crate settings.

Yep, I agree.

> Not sure if it makes much sense to keep the `FS_*` messages specifically, though...

Good point, if they are not used by anyone, we could remove them and add them back with a feature when they are needed.

germag:
> Not sure if it makes much sense to keep the FS_* messages specifically, though...

[...]

PR #246, opened 2024-06-27
vhost_user: Remove support for FS_* requests

aesteve-rh:
### Summary of the PR

Since these non-standard backend request
message types are confirmed not to be used
anymore, let's just remove them.

```
pub enum BackendReq {
    ...
    /// Virtio-fs draft: map file content into the window.
    FS_MAP = 100,
    FS_UNMAP = 1001,
    FS_SYNC = 1002,
    FS_IO = 1003,
    ...
}
```

Closes: #213

### Requirements

requirements:

  test.

### DIFF
--- a/vhost/CHANGELOG.md
+++ b/vhost/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### Changed
 
+### Remove
+- [[#246]](https://github.com/rust-vmm/vhost/pull/246) Remove support for FS_* requests
+
 ### Fixed
 
 ### Deprecated
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 72.12,
+  "coverage_score": 73.04,
   "exclude_path": "vhost/src/vhost_kern/",
   "crate_features": "vhost/vhost-user-frontend,vhost/vhost-user-backend,vhost-user-backend/postcopy"
 }
--- a/vhost/src/vhost_user/backend_req.rs
+++ b/vhost/src/vhost_user/backend_req.rs
@@ -181,16 +181,6 @@ impl VhostUserFrontendReqHandler for Backend {
             Some(&[fd.as_raw_fd()]),
         )
     }
-
-    /// Forward vhost-user-fs map file requests to the backend.
-    fn fs_backend_map(&self, fs: &VhostUserFSBackendMsg, fd: &dyn AsRawFd) -> HandlerResult<u64> {
-        self.send_message(BackendReq::FS_MAP, fs, Some(&[fd.as_raw_fd()]))
-    }
-
-    /// Forward vhost-user-fs unmap file requests to the frontend.
-    fn fs_backend_unmap(&self, fs: &VhostUserFSBackendMsg) -> HandlerResult<u64> {
-        self.send_message(BackendReq::FS_UNMAP, fs, None)
-    }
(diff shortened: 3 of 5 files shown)


aesteve-rh:
The best way to increase coverage, removing old code :)

vhost/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### Changed
 
+### Remove
+- [[#241]](https://github.com/rust-vmm/vhost/pull/246) Remove support for FS_* requests
```
stefano-garzarella, line 10:
```suggestion
- [[#246]](https://github.com/rust-vmm/vhost/pull/246) Remove support for FS_* requests
```

stefano-garzarella (commented):
Thanks for doing that :-) Just a minor typo, the rest LGTM.

germag:
LGTM

stefano-garzarella (approved):

germag (approved):

rbradford:
Unfortunately this change prevents Cloud Hypervisor from upgrading to the newer version of the vhost crate. Would it be possible to revert this until we can come up with a solution that allows us to support non-standard messages?

stefano-garzarella:
@rbradford oh, we were not aware about users, indeed we had #213 opened for a while about removing those.

IMHO we can reintroduce them since they are not in conflicts, but from commit 85e00973f4a1c77b4dfb8e4a4dde0489d944db61 we changed their value. Is your code aware of that?

@germag @aesteve-rh WDYT? If we reintroduce them, shoule we add them back behind a feature?

aesteve-rh:
@rbradford I think you could open a new Issue so that the request gets more visibility.

IMO if we add these back, you should take into account that:
- They will not get the same request ID they had originally, as those now overlap with the standard, they should be reverted to the 1000-1004 range. So it would require adapting the code from your side.
- Take into account that there is an effort in place to standardize these messages, with very similar semantics: https://lore.kernel.org/all/20240912144432.126717-1-aesteve@redhat.com/T/#t

rbradford:
I'm afraid i'm not a vhost-user or virtio-fs expert and the person who was most involved in that part is no longer as active on Cloud Hypervisor. It's possible i've misunderstood - are these commands not required for correct virtio-fs operation? (We only support running against the Rust virtiofsd implementation.)

rbradford:
> > I'm afraid i'm not a vhost-user or virtio-fs expert and the person who was most involved in that part is no longer as active on Cloud Hypervisor. It's possible i've misunderstood - are these commands not required for correct virtio-fs operation?
> 
> They are not in the spec, so my guess is that they are not required till your frontend/backend follow the spec.
> 
> > (We only support running against the Rust virtiofsd implementation.)
> 
> @germag is the maintainer of [virtiofsd](https://gitlab.com/virtio-fs/virtiofsd), and he approved this PR. We also discussed that in #213 where we received a confirmation that virtiofsd is not using them at all.

Cool - thank you for the clarification. It's possible it's legacy left over from when the virtio-fs support was added (I believe we did support DAX originally). Let me try removing this and see where we are.

stefano-garzarella:
> Cool - thank you for the clarification. It's possible it's legacy left over from when the virtio-fs support was added (I believe we did support DAX originally). Let me try removing this and see where we are.

@rbradford yeah, it could be.
If you find you need them, please follow @aesteve-rh suggestion and move the discussion to a new issue, so we can track it.

PR #251, opened 2024-08-27
vhost_user: Add support for SHMEM_MAP/UNMAP backend requests

mtjhrc:
### Summary of the PR

This PR adds support for backend requests `SHMEM_MAP` and `SHMEM_UNMAP`. 
This allows the device to memory map a file descriptor into a  `Shared Memory Region`.

The `SHMEM_MAP` and `SHMEM_UNMAP` are part of this QEMU RFC:
> [RFC PATCH v2 0/5] vhost-user: Add SHMEM_MAP/UNMAP requests
https://mail.gnu.org/archive/html/qemu-devel/2024-06/msg05736.html

I don't anticipate these specific requests to change much, the discussion for this RFC is mainly about a situation where one device is accessing shared memory mapped by another device.

These requests are required for implementing blob resource support in vhost-user-gpu. 
[The current gpu PR](
https://github.com/rust-vmm/vhost-device/pull/668
), doesn't implement blob resources and as such doesn't require this, but I am working on creating another PR with blob resource support.

This is based on a commit by Albert Esteve <aesteve@redhat.com> in his branch/fork.

### Requirements

requirements:

  test.


vhost/src/vhost_user/backend_req.rs
```diff
@@ -183,6 +183,16 @@ impl VhostUserFrontendReqHandler for Backend {
             Some(&[fd.as_raw_fd()]),
         )
     }
+
+    /// Forward vhost-user memory map file requests to the backend.
```
stefano-garzarella, line 187:
```suggestion
    /// Forward vhost-user memory map file requests to the fronted.
```

vhost/src/vhost_user/frontend_req_handler.rs
```diff
@@ -52,6 +52,15 @@ pub trait VhostUserFrontendReqHandler {
...
 
+    /// Handle shared memory region mapping requests.
+    fn shmem_map(&self, _req: &VhostUserMMap, _fd: &dyn AsRawFd) -> HandlerResult<u64> {
+        Err(std::io::Error::from_raw_os_error(libc::ENOSYS))
+    }
+
+    /// Handle shared memory region unmapping requests.
+    fn shmem_unmap(&self, _req: &VhostUserMMap) -> HandlerResult<u64> {
+        Err(std::io::Error::from_raw_os_error(libc::ENOSYS))
+    }
```
stefano-garzarella, line 63: I'd add a new line here

vhost/src/vhost_user/message.rs
```diff
@@ -194,6 +194,10 @@ enum_value! {
         SHARED_OBJECT_REMOVE = 7,
         /// Lookup for a virtio shared object.
         SHARED_OBJECT_LOOKUP = 8,
+        /// Map memory into guest address space ()
```
stefano-garzarella, line 197: The last few characters ` ()`  may be typos.

vhost/src/vhost_user/message.rs
```diff
@@ -990,6 +994,48 @@ impl VhostUserMsgValidator for VhostUserTransferDeviceState {
...
+    /// Struct padding.
+    pub padding: [u8; 7],
+    /// File offset.
+    pub fd_offset: u64,
+    /// Offset into the shared memory region.
+    pub shm_offset: u64,
+    /// Size of region to map.
+    pub len: u64,
+    /// Flags for the mmap operation
+    pub flags: VhostUserMMapFlags,
```
stefano-garzarella, line 1026: I'd follow what we did for other struct like `struct VhostUserMsgHeader` and `struct VhostUserHeaderFlag`. So here I'd use `u64`, also because in the `VhostUserMsgValidator` implementation you're already calling `.bits()`.

stefano-garzarella (commented):
Author and SoB are different, can you fix it, or add @aesteve-rh SoB if he is a co-author.

Please check also the coverage and add a line in the changelog.

mtjhrc:
I have addressed the feedback, you can take another look.

Also, should I add some comment that this feature is not really "stable"? Or will PR have to wait until it gets merged in QEMU?

mtjhrc:
The coverage was too low, so I wanted to add a test for the default impl of the frontend request handlers, but I also added coverage for all the methods, since that seemed quite trivial. This has increased the coverage too much, so I had to increase the coverage config value.

stefano-garzarella:
The code LGTM, but before merging this PR, I'd like to wait at least the spec merged in QEMU.
We recently removed some code that was never merged in the spec, so I'd like to avoid making the same mistake.
If the QEMU RFC still requires discussion about the code, maybe we can try to merge at least the changes to the specification so we can move forward here as well.

WDYT?

mtjhrc:
I understand the concern. I'm talking to @aesteve-rh to see if we can figure something out

vhost/src/vhost_user/message.rs
```diff
@@ -990,6 +994,48 @@ impl VhostUserMsgValidator for VhostUserTransferDeviceState {
     }
 }
 
+// Bit mask for flags in VhostUserBackendMapMsg backend message
+bitflags! {
+    #[derive(Copy, Clone, Debug, Eq, PartialEq, Default)]
+    /// Flags for virtio-fs backend messages.
```
alyssais, line 1000: virtio-fs?

mtjhrc: Oops, fixed it. It was a leftover from the removed virtio-fs DAX messages.

vhost/src/vhost_user/message.rs
```diff
@@ -990,6 +994,48 @@ impl VhostUserMsgValidator for VhostUserTransferDeviceState {
...
+
+// SAFETY: Safe because all fields of VhostUserBackendMapMsg are POD.
+unsafe impl ByteValued for VhostUserMMap {}
+
+impl VhostUserMsgValidator for VhostUserMMap {
+    fn is_valid(&self) -> bool {
+        (self.flags & !VhostUserMMapFlags::all().bits()) == 0
+            && self.fd_offset.checked_add(self.len).is_some()
+            && self.shm_offset.checked_add(self.len).is_some()
+    }
```
dorindabassey, line 1037: do you want to add more checks relevant to the struct? e.g a check based on valid values for `shmid`?

vhost/src/vhost_user/frontend_req_handler.rs
```diff
@@ -395,6 +437,16 @@ mod tests {
             }
             Ok(1)
         }
+
+        fn shmem_map(&mut self, req: &VhostUserMMap, _fd: &dyn AsRawFd) -> HandlerResult<u64> {
+            assert_eq!({ req.shmid }, 0);
+            Ok(!self.shmem_mappings.insert((req.shm_offset, req.len)) as u64)
```
dorindabassey, line 444: i think it's better to remove the negation (!) and use a direct check to return meaningful values. e.g `if self.shmem_mappings.insert(req.shm_offset, req.len) { Ok(1) } else { Ok(0) }` and why the double braces? missing &?

vhost/src/vhost_user/frontend_req_handler.rs
```diff
@@ -395,6 +437,16 @@ mod tests {
...
+        }
+
+        fn shmem_unmap(&mut self, req: &VhostUserMMap) -> HandlerResult<u64> {
+            assert_eq!({ req.shmid }, 0);
+            Ok(!self.shmem_mappings.remove(&(req.shm_offset, req.len)) as u64)
```
dorindabassey, line 449: same here

vhost/src/vhost_user/backend_req.rs
```diff
@@ -191,10 +191,16 @@ mod tests {
 
     use super::*;
 
+    fn frontend_backend_pair() -> (Endpoint<VhostUserMsgHeader<BackendReq>>, Backend) {
+        let (p1, p2) = UnixStream::pair().unwrap();
```
dorindabassey, line 205: Instead of using .unwrap() do you want to handle the error here?

mtjhrc: Well it's utility function only used inside tests, I don't think the error needs to be forwarded here. I think the only situation where this can fail is when maximum file descriptor limit is reached.

dorindabassey:
> The code LGTM, but before merging this PR, I'd like to wait at least the spec merged in QEMU. We recently removed some code that was never merged in the spec, so I'd like to avoid making the same mistake. If the QEMU RFC still requires discussion about the code, maybe we can try to merge at least the changes to the specification so we can move forward here as well.
> 
> WDYT?

are you talking about the vhost-user specification https://qemu-project.gitlab.io/qemu/interop/vhost-user.html#message-specification?  
If so, I think it make sense to atleast have just the spec merged, since the patches are still RFC it might take a while before it's merged.

vhost/src/vhost_user/frontend_req_handler.rs
mtjhrc: By braces do you mean `( )`? The `(req.shm_offset, req.len)` is a tuple.

In regards to the negation with cast instead of explicit `if`, I considered to write it like you did, I mainly wrote it like this, because it is symmetric with the existing `shared_object_add`, `shared_object_remove` above:
```Rust
fn shared_object_add(&mut self, uuid: &VhostUserSharedMsg) -> HandlerResult<u64> {
    Ok(!self.shared_objects.insert(uuid.uuid) as u64)
}
fn shared_object_remove(&mut self, uuid: &VhostUserSharedMsg) -> HandlerResult<u64> {
    Ok(!self.shared_objects.remove(&uuid.uuid) as u64)
}
```

I am not sure if I should change it.

dorindabassey: >  The (req.shm_offset, req.len) is a tuple.

ah right!

> because it is symmetric with the existing shared_object_add, shared_object_remove above:

doesn't necessarily mean it's the correct approach.

mtjhrc: The `{ }` braces are required here to force the value to be copied. Without the braces this fails to compile because  `assert_eq!` macro tries to take a reference to `req.shmid` which is not possible. This is because the struct `VhostUserMMap` is `repr(packed)` so it is `unsafe` to take references to its fields (they may not be aligned).

vhost/src/vhost_user/message.rs
mtjhrc: From virtio spec:
> A device may have multiple shared memory regions associated with it. Each region has a shmid to identify it, the meaning of which is device-specific. 

According to the `virtio` (not vhost) spec the valid values of `shmid` are dependent on the device type (and the advertised device features), so I don't think you could check it here generally.

stefano-garzarella:
> > The code LGTM, but before merging this PR, I'd like to wait at least the spec merged in QEMU. We recently removed some code that was never merged in the spec, so I'd like to avoid making the same mistake. If the QEMU RFC still requires discussion about the code, maybe we can try to merge at least the changes to the specification so we can move forward here as well.
> > WDYT?
> 
> are you talking about the vhost-user specification https://qemu-project.gitlab.io/qemu/interop/vhost-user.html#message-specification? 

@dorindabassey yep

> If so, I think it make sense to atleast have just the spec merged, since the patches are still RFC it might take a while before it's merged.

yeah, if it's possible, I'd go in that direction.
So, I'm marking this PR as draft to be sure we don't merge this till we have the spec changes merged in QEMU.

vhost/src/vhost_user/message.rs
dorindabassey: ok, fair enough for `shmid`, you can leave it as it is.

vhost/src/vhost_user/message.rs
```diff
@@ -990,6 +994,48 @@ impl VhostUserMsgValidator for VhostUserTransferDeviceState {
...
+    #[derive(Copy, Clone, Debug, Eq, PartialEq, Default)]
+    /// Flags specifying access permissions of memory mapping of a file
+    pub struct VhostUserMMapFlags: u64 {
+        /// Empty permission.
+        const EMPTY = 0x0;
+        /// Read permission.
+        const MAP_R = 0x1;
+        /// Write permission.
+        const MAP_W = 0x2;
+    }
```
aesteve-rh, line 1005: Flags changed, they are now (as seen in the latest version of the spec):
```
:flags: a 64-bit value:
  - 0: Pages are mapped read-only
  - 1: Pages are mapped read-write
```

vhost/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 
 ### Added
 - [[#268]](https://github.com/rust-vmm/vhost/pull/268) Add support for `VHOST_USER_GET_SHARED_OBJECT`
+- [[#251]](https://github.com/rust-vmm/vhost/pull/251) Add `SHMEM_MAP` and `SHMEM_UNMAP` support
```
epilys: This should go in the "Unreleased - Added" section above

vhost/src/vhost_user/backend_req.rs
```diff
@@ -257,4 +270,50 @@ mod tests {
             .shared_object_add(&VhostUserSharedMsg::default())
             .unwrap();
     }
+
+    #[test]
+    fn test_shmem_map() {
+        let (mut fronted, backend) = frontend_backend_pair();
```
epilys:
```suggestion
        let (mut frontend, backend) = frontend_backend_pair();
```

vhost/src/vhost_user/message.rs
```diff
@@ -987,6 +991,45 @@ impl VhostUserMsgValidator for VhostUserTransferDeviceState {
...
+    pub fd_offset: u64,
+    /// Offset into the shared memory region.
+    pub shm_offset: u64,
+    /// Size of region to map.
+    pub len: u64,
+    /// Flags for the mmap operation
+    pub flags: u64,
+}
+
+// SAFETY: Safe because all fields of VhostUserBackendMapMsg are POD.
```
dorindabassey: Typo? I don't see any `VhostUserBackendMapMsg` struct

mtjhrc: Fixed - the struct got renamed (it should refer to `VhostUserMMap`)

vhost/src/vhost_user/message.rs
```diff
@@ -987,6 +991,45 @@ impl VhostUserMsgValidator for VhostUserTransferDeviceState {
     }
 }
 
+// Bit mask for flags in VhostUserMMap struct
+bitflags! {
+    #[derive(Copy, Clone, Debug, Eq, PartialEq, Default)]
+    /// Flags specifying access permissions of memory mapping of a file
+    /// Use VhostUserMMapFlags::default() for a read-only mapping
+    pub struct VhostUserMMapFlags: u64 {
+        /// Read-write permission
```
stefano-garzarella, line 1003: nit: `/// Read-write permission (0: read-only, 1: read-write)`

vhost/src/vhost_user/backend_req.rs
```diff
@@ -174,6 +174,16 @@ impl VhostUserFrontendReqHandler for Backend {
             Some(&[fd.as_raw_fd()]),
         )
     }
+
+    /// Forward vhost-user memory map file request to the frontend.
+    fn shmem_map(&self, req: &VhostUserMMap, fd: &dyn AsRawFd) -> HandlerResult<u64> {
+        self.send_message(BackendReq::SHMEM_MAP, req, Some(&[fd.as_raw_fd()]))
```
stefano-garzarella, line 195: In `shared_object_*()` functions above we check if the related feature is negotiated, should we do something similar also here?

vhost/src/vhost_user/message.rs
stefano-garzarella, line 1: I guess we also need to add `SHMEM` to `VhostUserProtocolFeatures` as defined in https://lore.kernel.org/all/20251111091058.879669-4-aesteve@redhat.com/:

```diff
@@ -1064,6 +1086,7 @@ Protocol features
   #define VHOST_USER_PROTOCOL_F_XEN_MMAP             17
   #define VHOST_USER_PROTOCOL_F_SHARED_OBJECT        18
   #define VHOST_USER_PROTOCOL_F_DEVICE_STATE         19
+  #define VHOST_USER_PROTOCOL_F_SHMEM                20
 
 Front-end message types
 -----------------------
```

vhost/src/vhost_user/message.rs
```diff
@@ -987,6 +991,45 @@ impl VhostUserMsgValidator for VhostUserTransferDeviceState {
...
+    /// Size of region to map.
+    pub len: u64,
+    /// Flags for the mmap operation
+    pub flags: u64,
+}
+
+// SAFETY: Safe because all fields of VhostUserMMap are POD.
+unsafe impl ByteValued for VhostUserMMap {}
+
+impl VhostUserMsgValidator for VhostUserMMap {
```
stefano-garzarella, line 1029: Should we add a test for this, as we do for other `VhostUserMsgValidator`  implementations (e.g. `check_user_memory()`, `check_user_memory_region()`, etc.)

vhost/src/vhost_user/message.rs
```diff
@@ -987,6 +991,45 @@ impl VhostUserMsgValidator for VhostUserTransferDeviceState {
...
+    fn is_valid(&self) -> bool {
+        VhostUserMMapFlags::from_bits(self.flags).is_some()
```
stefano-garzarella: Is `self.len == 0` valid?

stefano-garzarella (changes requested):

vhost/src/vhost_user/message.rs
stefano-garzarella: Should we also check if `padding` is set to 0 like we do in `VhostUserMemory` ?
This is more for the future, where we can re-use the padding for something else. That said, in the spec I don't see any push on that :-(

vhost/src/vhost_user/message.rs
epilys: The comment is for `WRITABLE` constant, which is both read-write. Maybe `(0: read-only, 1: read-write)` could go in the struct doc?

stefano-garzarella: ah, I thought was a bitfields.

Or maybe `/// Read-write permission (if not set, it's read-only)`, but I see we already have something in the struct doc, so really not a strong opinion on this, feel free to do nothing ;-)

vhost/src/vhost_user/message.rs
epilys: > Is self.len == 0 valid?

It's not mentioned in the spec text, but in the current patches in QEMU it's rejected:

```diff
+VirtioSharedMemoryMapping *virtio_shared_memory_mapping_new(uint8_t shmid,
+                                                            int fd,
+                                                            uint64_t fd_offset,
+                                                            uint64_t shm_offset,
+                                                            uint64_t len,
+                                                            bool allow_write)
+{
+    VirtioSharedMemoryMapping *mapping;
+    MemoryRegion *mr;
+    g_autoptr(GString) mr_name = g_string_new(NULL);
+    uint32_t ram_flags;
+    Error *local_err = NULL;
+
+    if (len == 0) {
+        error_report("Shared memory mapping size cannot be zero");
+        return NULL;
+    }
+
```

> Should we also check if padding is set to 0 like we do in VhostUserMemory ?

Sounds too strict IMHO... what if padding is filled with garbage bytes but the rest of the message is valid?

stefano-garzarella: IMO in the spec we should say that padding must be 0, so in the future we can potentially use it for something else. At least, this is what we usually do in Linux uAPI.

stefano-garzarella: > IMO in the spec we should say that padding must be 0, so in the future we can potentially use it for something else. At least, this is what we usually do in Linux uAPI.

Just to be clear, this is not a blocker for this PR ;-)

aesteve-rh: > IMO in the spec we should say that padding must be 0, so in the future we can potentially use it for something else. At least, this is what we usually do in Linux uAPI.

That's a good point. Honestly, I did not think about it. I will send a follow-up patch (I have a few minor changes piled already...) once it lands on QEMU to complete the spec with both points raised here.

Perhaps we can add the check for `len==0` here and wait for the padding check for when I can send the follow-up.

epilys: And make it `NonZero<_>` while we're at it...

stefano-garzarella: @aesteve-rh thanks for that. If it's supposed to go into the spec, IMHO we can implement that already here. In the worst case, we can relax the requirement here if the spec will not require that.

germag: sorry if I'm misunderstanding this, I'm late for this, but uninit padding is not ok, it's a source of UB in rust, specially if the struct is `ByteValued`, we already had that problem (and padding is not the same that volatile memory)

germag: I think I'm also late to ask for `packed`  in the spec/qemu... I have in my TODO to fix a couple of struct because of that

vhost/src/vhost_user/message.rs
mtjhrc: I mean what's best here, depends how the protocol evolves. I could theoretically imagine it becoming a real bitfield in the future with more options.

vhost/src/vhost_user/message.rs
aesteve-rh: @germag for the spec part, would that be ok if I address that in a follow up? The patch has not been merged but after how long it took to have it ready, I do not want to cause delays. I have a few other comments to address, so I will make sure to send it and put you on cc.

germag: > @germag for the spec part, would that be ok if I address that in a follow up? The patch has not been merged but after how long it took to have it ready, I do not want to cause delays. I have a few other comments to address, so I will make sure to send it and put you on cc.

Yes, I think it's fair, it was my fault not reviewing it on time, thanks!

mtjhrc: @germag 
> sorry if I'm misunderstanding this, I'm late for this, but uninit padding is not ok, it's a source of UB in rust, specially if the struct is `ByteValued`, we already had that problem (and padding is not the same that volatile memory)

I don't see this being a problem here, we are not constructing uninitialized memory on the Rust side. The memory could be uninitiated on the sender side, but the sender writes the unilized bytes to a socket, and we override the whole `ByteValued` struct with the bytes received from the socket, hence we *do* initialize it (potentially to some unknown bytes from the socket though).

mtjhrc: @epilys  
> And make it NonZero<_> while we're at it...

I can't  change the type to type of `len` field to  `NonZero<u64>` in `VhostUserMMap`, because that would cause UB when other party sends us `0` in the `len` field. 
(I could change it to `Option<NonZero<u64>`, which should be FFI compatible with `u64`, but that won't really help anything)

I mean we kind of need two msg structs for each msg, one struct for over the wire protocol, that **can** contain invalid msgs, and one for high level usage that cannot, doesn't have useless fields like padding, uses fancy Option, NonZero... types.  (or just have one struct and make the fields private, add a safe builder + getters...). I think this is a more general problem that should probably be addressed more globally.

epilys: @mtjhrc good point. It's unnecessary, validation is fine also.

vhost/src/vhost_user/message.rs
mtjhrc: Right, I missed that, done.

vhost/src/vhost_user/backend_req.rs
mtjhrc: Done

vhost/src/vhost_user/message.rs
mtjhrc: Makes sense, I've added a test

vhost/src/vhost_user/message.rs
mtjhrc: I added the check that verifies that `len` cannot be zero. I also added a comment that we don't validate the padding because it's currently not specified in the spec.

vhost/src/vhost_user/message.rs
stefano-garzarella: @mtjhrc yeah, I agree, let's postpone this.

epilys (dismissed):

vhost/src/vhost_user/backend_req.rs
```diff
@@ -174,6 +186,22 @@ impl VhostUserFrontendReqHandler for Backend {
             Some(&[fd.as_raw_fd()]),
         )
     }
+
+    /// Forward vhost-user memory map file request to the frontend.
+    fn shmem_map(&self, req: &VhostUserMMap, fd: &dyn AsRawFd) -> HandlerResult<u64> {
+        if !self.node().shmem_negotiated {
+            return Err(io::Error::other("SHMEM feature not negotiated"));
+        }
+        self.send_message(BackendReq::SHMEM_MAP, req, Some(&[fd.as_raw_fd()]))
```
stefano-garzarella, line 195: This is pre-existing in the other handlers, so not blocking for this PR, but here IMO we should take the lock guard (`let node = self.node()`) at the beginning, and use that guard for both feature checking and message sending.

Also not sure why we called it node if it's a guard, but yeah...

vhost/src/vhost_user/frontend_req_handler.rs
stefano-garzarella, line 1: nit: s/implemenation/implementation in the commit https://github.com/rust-vmm/vhost/pull/251/commits/f2a0c2b2d2ffd912e5f2def497d4acc764c7998c description

vhost/src/vhost_user/backend_req.rs
```diff
@@ -257,4 +288,60 @@ mod tests {
...
+            .unwrap_err();
+
+        backend.set_shmem_flag(true);
+        backend.shmem_map(&map_request, &some_fd_to_send).unwrap();
+
+        let (hdr, request, fd) = frontend.recv_body::<VhostUserMMap>().unwrap();
+        assert_eq!(hdr.get_code().unwrap(), BackendReq::SHMEM_MAP);
+        assert!(fd.is_some());
+        assert_eq!({ request.shm_offset }, { map_request.shm_offset });
+        assert_eq!({ request.len }, { map_request.len },);
```
stefano-garzarella: nit:

```suggestion
        assert_eq!({ request.len }, { map_request.len });
```

stefano-garzarella: Also, why we need `{` `}`  ?

vhost/src/vhost_user/backend_req.rs
stefano-garzarella: BTW I can send a PR later to fix all of those.

vhost/src/vhost_user/backend_req.rs
epilys: rustfmt doesn't format inside macros unfortunately, so trailing commas and unnecessary blocks pass through

mtjhrc: When, I looked at this after I while, I was as puzzled as you.

But when I remove it:
```
error[E0793]: reference to packed field is unaligned
   --> vhost/src/vhost_user/backend_req.rs:318:9
    |
318 |         assert_eq!(request.len, map_request.len);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: packed structs are only aligned by one byte, and many modern architectures penalize unaligned field accesses
    = note: creating a misaligned reference is undefined behavior (even if that reference is never dereferenced)
    = help: copy the field contents to a local variable, or replace the reference with a raw pointer and use `read_unaligned`/`write_unaligned` (loads and stores via `*p` must be properly aligned even when using raw pointers)
    = note: this error originates in the macro `assert_eq` (in Nightly builds, run with -Z macro-backtrace for more info)

```

The extra `{` `}` block forces a Copy avoiding the usage of a reference.

stefano-garzarella: @mtjhrc oh, I see, okay, makes sense!

epilys (approved):

PR #252, opened 2024-08-28
Fix bug in unsafe code of test create_listener_from_raw_fd()

mtjhrc:
### Summary of the PR
This is a small PR to fix a bug in test `create_listener_from_raw_fd()`, to make sure a file descriptor is not closed after ownership is passed to another object.

This actually causes a crash when running `cargo test` using the latest Rust toolchain version `1.80`, with the error message:
```
fatal runtime error: IO Safety violation: owned file descriptor already closed
```
 

### Requirements

requirements:

  test.

### DIFF
--- a/vhost/src/vhost_user/connection.rs
+++ b/vhost/src/vhost_user/connection.rs
@@ -617,6 +617,7 @@ fn get_sub_iovs_offset(iov_lens: &[usize], skip_size: usize) -> (usize, usize) {
 mod tests {
     use super::*;
     use std::io::{Read, Seek, SeekFrom, Write};
+    use std::os::fd::IntoRawFd;
     use vmm_sys_util::rand::rand_alphanumerics;
     use vmm_sys_util::tempfile::TempFile;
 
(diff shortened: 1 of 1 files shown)


mtjhrc:
Should I add this to the changelog too? This doesn't fix user facing functionality, only fixes the test.

stefano-garzarella (approved):

stefano-garzarella:
> Should I add this to the changelog too? This doesn't fix user facing functionality, only fixes the test.

I agree that we can skip a changelog entry. Thanks for the fix!

alyssais (approved):

germag (approved):

germag:
I think we should also check the `eventfd.as_raw_fd()` usage in `vhost_user_backend`, maybe `vmm_sys_util::linux::eventfd::EventFd` should implement `IntoRawFd`.

But, I really prefer swap `File`/ `RawFd` to `OwnedFd`/`BorrowedFd` removing the unsafe from the API and doing the conversion internally

stefano-garzarella:
> I think we should also check the `eventfd.as_raw_fd()` usage in `vhost_user_backend`, maybe `vmm_sys_util::linux::eventfd::EventFd` should implement `IntoRawFd`.
> 
> But, I really prefer swap `File`/ `RawFd` to `OwnedFd`/`BorrowedFd` removing the unsafe from the API and doing the conversion internally

Agree on that!

PR #260, opened 2024-10-02
vhost-user-backend: impl Debug for AddrMapping

tylerfanelli:
### Summary of the PR

It can be useful to debug the various address mappings sent to a vhost user device backend. impl Debug for both the postcopy and non-postcopy versions of AddrMapping.

### DIFF
--- a/vhost-user-backend/src/handler.rs
+++ b/vhost-user-backend/src/handler.rs
@@ -82,6 +82,7 @@ impl error::Error for VhostUserHandlerError {}
 /// Result of vhost-user handler operations.
 pub type VhostUserHandlerResult<T> = std::result::Result<T, VhostUserHandlerError>;
 
+#[derive(Debug)]
 struct AddrMapping {
     #[cfg(feature = "postcopy")]
     local_addr: u64,


vhost-user-backend/src/handler.rs
```diff
@@ -90,6 +91,29 @@ struct AddrMapping {
...
+impl fmt::Debug for AddrMapping {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("AddrMapping")
+            .field("vmm_addr", &format!("0x{:x}", &self.vmm_addr))
+            .field("size", &format!("0x{:x}", &self.size))
+            .field("gpa_base", &format!("0x{:x}", &self.gpa_base))
+            .finish()
+    }
+}
+
```
germag, line 116: Why not just?:

```rust
#[derive(Debug)]
struct AddrMapping {
...
}
```

since all `AddrMapping`'s fields are primitive

tylerfanelli: Addresses are not represented as hexadecimal with `#[derive(Debug)]`

rbradford: If the caller wants hex values they can do "{:x?}"

tylerfanelli: That's news to me :) thanks for showing that. In that case `#[derive(Debug)]` is sufficient. Will make the change.

germag (approved):

stefano-garzarella (approved):

Issue #265, opened 2024-10-16
gpu-socket feature causing feature unification issues


dorindabassey:
since the set_gpu_socket feature is behind a feature flag, this cause feature unification issues, because vhost-device-gpu is enabling that feature and other crates are not enabling it. According to rust feature reference: 

> That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. A feature should not introduce a [SemVer-incompatible change](https://doc.rust-lang.org/cargo/reference/features.html#semver-compatibility).

dorindabassey:
@stefano-garzarella is suggesting that we remove the `gpu_socket` feature flag. which I agree can solve the feature unification problem.
more suggestions are welcomed.

stefano-garzarella:
Just a bit more context. The issue was discovered while adding `vhost-user-gpu` crate in a workspace with other crates using the `vhost-user-backend` crate, see https://github.com/rust-vmm/vhost-device/pull/668

Since the new crate enable `gpu-socket` feature, this is enabled by cargo also for other crates in the workspace to minimize the number of copy of the crate, see https://doc.rust-lang.org/cargo/reference/features.html#feature-unification

At this point, we had the following issue on another crate in the same workspace (`vhost-device-console`) not enabling that feature:
```
   Compiling vhost-device-console v0.1.0 (/workdir/staging/vhost-device-console)
error[E0046]: not all trait items implemented, missing: `set_gpu_socket`
   --> vhost-device-console/src/vhu_console.rs:641:1
    |
641 | impl VhostUserBackendMut for VhostUserConsoleBackend {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `set_gpu_socket` in implementation
```

The first fix we had in mind was in some way to avoid the feature unification, maybe forcing cargo to use 2 copies, but at that point we discovered the recommendation in https://doc.rust-lang.org/cargo/reference/features.html#feature-unification that also @dorindabassey reported:

> That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. A feature should not introduce a [SemVer-incompatible change](https://doc.rust-lang.org/cargo/reference/features.html#semver-compatibility).

So maybe we should avoid that here, so I was thinking of not really removing the feature, but something like just providing default implementation for VhostUserBackend that should not be used if the feature is not enabled (maybe we can panic there):
```diff
[...]

PR #266, opened 2024-10-16
Add suport for VHOST_USER_RESET_DEVICE

XanClic:
### Summary of the PR

The main point of this PR is to add vhost-user back-end support for the `RESET_DEVICE` message. This message is handled by disabling all vrings and resetting the device state (as per [the specification, look for `VHOST_USER_RESET_DEVICE`](https://qemu-project.gitlab.io/qemu/interop/vhost-user.html#front-end-message-types)). This includes a new `VhostUserBackend::reset_device()` method that back-end implementations may provide to clear additional internal state they keep.

virtiofsd needs this so a hard reset cannot be used to inherit state from a prior user ([virtiofsd issue #167](https://gitlab.com/virtio-fs/virtiofsd/-/issues/167) – note that the issue describes a bug in qemu that would prevent it from sending `RESET_DEVICE`, which has [since been fixed](https://gitlab.com/qemu-project/qemu/-/commit/2688e8df60f5a655dc34c5e38523e425556f8483)).

Like all other protocol features, it is left to the back-end’s `VhostUserBackend::protocol_features()` implementation to decide whether to announce support for this or not.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-user-backend/src/backend.rs
+++ b/vhost-user-backend/src/backend.rs
@@ -58,6 +58,12 @@ pub trait VhostUserBackend: Send + Sync {
     /// Get available vhost protocol features.
     fn protocol_features(&self) -> VhostUserProtocolFeatures;
 
+    /// Reset the emulated device state.
+    ///
+    /// A default implementation is provided as we cannot expect all backends to implement this
+    /// function.
+    fn reset_device(&self) {}
+
     /// Enable or disable the virtio EVENT_IDX feature
     fn set_event_idx(&self, enabled: bool);
 
@@ -179,6 +185,12 @@ pub trait VhostUserBackendMut: Send + Sync {
 
     /// Get available vhost protocol features.
     fn protocol_features(&self) -> VhostUserProtocolFeatures;
+
+    /// Reset the emulated device state.
+    ///
+    /// A default implementation is provided as we cannot expect all backends to implement this
+    /// function.
+    fn reset_device(&mut self) {}
 
     /// Enable or disable the virtio EVENT_IDX feature
     fn set_event_idx(&mut self, enabled: bool);
@@ -305,6 +317,10 @@ impl<T: VhostUserBackend> VhostUserBackend for Arc<T> {
         self.deref().protocol_features()
     }
 
+    fn reset_device(&self) {
+        self.deref().reset_device()
+    }
+
     fn set_event_idx(&self, enabled: bool) {
         self.deref().set_event_idx(enabled)
     }
@@ -384,6 +400,10 @@ impl<T: VhostUserBackendMut> VhostUserBackend for Mutex<T> {
 
     fn protocol_features(&self) -> VhostUserProtocolFeatures {
         self.lock().unwrap().protocol_features()
+    }
+
+    fn reset_device(&self) {
+        self.lock().unwrap().reset_device()
     }
 
     fn set_event_idx(&self, enabled: bool) {
@@ -468,6 +488,10 @@ impl<T: VhostUserBackendMut> VhostUserBackend for RwLock<T> {
 
     fn protocol_features(&self) -> VhostUserProtocolFeatures {
         self.read().unwrap().protocol_features()
+    }
+
+    fn reset_device(&self) {
+        self.write().unwrap().reset_device()
     }
 
     fn set_event_idx(&self, enabled: bool) {
@@ -589,6 +613,12 @@ pub mod tests {
             VhostUserProtocolFeatures::all()
         }
 
+        fn reset_device(&mut self) {
+            self.event_idx = false;
+            self.events = 0;
+            self.acked_features = 0;
+        }
+
         fn set_event_idx(&mut self, enabled: bool) {
             self.event_idx = enabled;
         }
@@ -668,6 +698,11 @@ pub mod tests {
             GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0x100000), 0x10000)]).unwrap(),
         );
         backend.update_memory(mem).unwrap();
+
+        backend.reset_device();
+        assert!(backend.lock().unwrap().events == 0);
+        assert!(!backend.lock().unwrap().event_idx);
+        assert!(backend.lock().unwrap().acked_features == 0);
     }
 
     #[test]
@@ -703,5 +738,10 @@ pub mod tests {
         backend
             .handle_event(0x1, EventSet::IN, &[vring], 0)
             .unwrap();
+
+        backend.reset_device();
+        assert!(backend.read().unwrap().events == 0);
+        assert!(!backend.read().unwrap().event_idx);
+        assert!(backend.read().unwrap().acked_features == 0);
     }
 }
--- a/vhost-user-backend/src/handler.rs
+++ b/vhost-user-backend/src/handler.rs
@@ -256,6 +256,19 @@ where
         Ok(())
     }
 
+    fn reset_device(&mut self) -> VhostUserResult<()> {
+        // Disable all vrings
+        for vring in self.vrings.iter_mut() {
+            vring.set_enabled(false);
+        }
+
+        // Reset device state, retain protocol state
+        self.features_acked = false;
+        self.acked_features = 0;
+        self.backend.reset_device();
+        Ok(())
+    }
+
     fn get_features(&mut self) -> VhostUserResult<u64> {
         Ok(self.backend.features())
     }
--- a/vhost-user-backend/tests/vhost-user-server.rs
+++ b/vhost-user-backend/tests/vhost-user-server.rs
@@ -61,6 +61,12 @@ impl VhostUserBackendMut for MockVhostBackend {
         VhostUserProtocolFeatures::all()
     }
 
+    fn reset_device(&mut self) {
+        self.events = 0;
+        self.event_idx = false;
+        self.acked_features = 0;
+    }
+
     fn set_event_idx(&mut self, enabled: bool) {
         self.event_idx = enabled;
     }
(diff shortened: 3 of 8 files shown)


XanClic:
First update already: Fixed the `CHANGELOG.md` entries to point to this MR (#266) instead of #265. (The latter came in just while I was creating this PR, and I also actually didn’t know that github uses the same namespace for both issues and PRs.)

XanClic:
v3: Fixed build failure in *vhost_kern*. I forgot to implement `reset_device()` there, but since there is no such ioctl, it’ll have to always return an error.

(And evidently I have a really hard time adding all the options to clippy that are necessary to cover everything…)

Ablu (approved):

vhost-user-backend/tests/vhost-user-server.rs
```diff
@@ -328,3 +354,38 @@ fn test_vhost_user_postcopy() {
...
+    barrier.wait();
+}
+
+#[test]
+fn test_vhost_user_reset_device() {
+    vhost_user_server_with_fn(vhost_user_reset_device, |backend, barrier| {
+        // Wait until `vhost_user_reset_device()` is before reset
+        barrier.wait();
+        // Check non-reset state
+        assert!(backend.lock().unwrap().acked_features == 0xffff_ffff_ffff_ffff);
```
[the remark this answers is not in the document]
XanClic: Sounds like a good idea! How about adding an associated `const FEATURES: u64` to `MockVhostBackend`?

vhost-user-backend/src/backend.rs
```diff
@@ -667,6 +691,11 @@ pub mod tests {
             GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0x100000), 0x10000)]).unwrap(),
         );
         backend.update_memory(mem).unwrap();
+
+        backend.reset_device();
+        assert!(backend.lock().unwrap().events == 0);
+        assert!(!backend.lock().unwrap().event_idx);
+        assert!(backend.lock().unwrap().acked_features == 0);
```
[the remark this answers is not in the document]
XanClic: True. After some consideration, I think I’d rather leave it as-is, though: Given that this is a test, not a hot path, execution time is not a problem, but more the fact that `.lock().unwrap()` clutters the code visually (I think).

We are at the end of the scope here, so in theory we could just lock `backend`, but I wouldn’t find that good style, because if at some point code is added afterwards, it will need to take care to limit the lock’s scope again. So we should rather limit the scope here, too.

But if I add a new scope, e.g. like
```rust
{
    let backend = backend.lock().unwap();
    assert!(backend.events == 0);
    ...
}
```
or
```rust
assert!({
    let backend = backend.lock().unwrap();
    backend.events == 0 && !backend.event_idx && backend.acked_features == 0
})
```
I don’t really find that more readable than just taking the lock every time anymore.

(Of course, I also wouldn’t mind, though, if you find any of those more pleasant to read. :slightly_smiling_face:)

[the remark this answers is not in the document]
Ablu: > We are at the end of the scope here, so in theory we could just lock backend, but I wouldn’t find that good style, because if at some point code is added afterwards, it will need to take care to limit the lock’s scope again.

I was thinking about doing it that way. But I do not mind too much (I approved after all :P). Happy to leave it is as it is.

vhost-user-backend/tests/vhost-user-server.rs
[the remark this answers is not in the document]
Ablu: Sounds good to me, maybe "REQUESTED_FEATURES" would make it slightly less generic... But I do not have any particular preference. I was only wondering about the meaning of the magic value (until I saw it being the same as above in `fn features()`).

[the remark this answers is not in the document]
XanClic: OK, I think I’ll make it `SUPPORTED_FEATURES`, then, because that value is what the mock “device” announces to be the features it can provide.

(`REQUESTED` sounds more like the ones the “guest driver” then actually requests to use, i.e. what `.acked_features` contains. It’s true that this value is really the one we should be verifying against, but that both depends on the device’s supported features and what the driver selects – I think it’s fair to rely on `setup_frontend()` to always request all of the provided feature bits.)

vhost-user-backend/src/backend.rs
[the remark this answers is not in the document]
XanClic: Got it!

Ablu (approved):

vhost-user-backend/src/backend.rs
```diff
@@ -177,6 +180,9 @@ pub trait VhostUserBackendMut: Send + Sync {
     /// Get available vhost protocol features.
     fn protocol_features(&self) -> VhostUserProtocolFeatures;
 
+    /// Reset the emulated device state.
+    fn reset_device(&mut self) {}
```
stefano-garzarella, line 184: Might it make sense to return an error if the backend-specific implementation for some reason does not support reset?

Although maybe just managing the feature is enough, at which point the backend doesn't enable it, also because from the specification it doesn't seem to be expected to fail.

vhost-user-backend/src/handler.rs
```diff
@@ -257,6 +257,19 @@ where
         Ok(())
     }
 
+    fn reset_device(&mut self) -> VhostUserResult<()> {
+        // Disable all vrings
```
stefano-garzarella, line 261: Should we check if `VHOST_USER_PROTOCOL_F_RESET_DEVICE` is negotiated or at least exposed by the backend?

The spec says:
```
``VHOST_USER_RESET_DEVICE``
...
  Only valid if the ``VHOST_USER_PROTOCOL_F_RESET_DEVICE`` protocol
  feature is set by the back-end.
```

vhost/src/backend.rs
```diff
@@ -254,6 +254,9 @@ pub trait VhostBackend: std::marker::Sized {
     /// This is no longer used.
     fn reset_owner(&self) -> Result<()>;
 
+    /// Disable all rings and reset the internal device state.
+    fn reset_device(&self) -> Result<()>;
```
stefano-garzarella, line 258: IIUC this is supported only by vhost-user, so what about adding it just in `VhostUserFrontend` ?

vhost-user-backend/src/backend.rs
XanClic: I think managing the feature is enough. Beyond that, I think reset should be an infallible operation; so if it’s generally supported, it should never fail in such a way where you’d return an error to qemu.

vhost-user-backend/src/handler.rs
XanClic: You mean adding a `check_protocol_feature()` function like `check_feature()`, and then run `self.check_protocol_feature(VhostUserProtocolFeatures::RESET_DEVICE)?;`? I don’t think other handlers do that, yet (e.g. `get_config()` should check `VHOST_USER_PROTOCOL_F_CONFIG`, but sure.

vhost/src/backend.rs
XanClic: That makes sense, yes. Will do.

vhost-user-backend/src/handler.rs
XanClic: Ah, `check_proto_feature()` it is, that’s how `FrontendInternal` calls it.

XanClic:
v5:
- Patch 1: Add `VhostUserHandler::check_proto_features()`, make use of it to gate `<VhostUserHandler as VhostUserBackendReqHandlerMut>::reset_device()` to error out when the `RESET_DEVICE` protocol feature has not been negotiated
- Patch 3: Move `reset_device()` from `VhostBackend` into `VhostUserFrontend`, given it’s only available on vhost-user anyway

vhost-user-backend/src/handler.rs
stefano-garzarella: Oh, I was checking `get_config()` and I noticed that we do the check in `BackendReqHandler::handle_request()` so now I'm thinking if it makes more sense to do it there, where we already have `check_proto_feature`.

Sorry for the confusion :-(

stefano-garzarella: That said, IMHO is better what we are doing here, so not sure if in the future we should move all the checks here.

So I'm fine with both this version or just move the check in the handle_request(). up to you ;-)

XanClic: Oh, right! I’ll move it there.

(The only other thing that checks feature availability in `impl VhostUserBackendReqHandlerMut for VhostUserHandler` is `fn set_vring_enable()`, which… also does it in `handle_request()`. Double-checking isn’t bad by any means, but that does show everyone expects there to be a check in `handle_request()`, and I think it makes sense to follow the majority of implementations and only check there then.)

XanClic:
v6: Moved the `F_RESET_DEVICE` feature gate check from `<VhostUserHandler as VhostUserBackendReqHandlerMut>::reset_device()` into `BackendReqHandler::handle_request()`, where all other request handlers have theirs.

stefano-garzarella (dismissed):
LGTM! Thanks!

vhost-user-backend/src/backend.rs
```diff
@@ -59,6 +59,9 @@ pub trait VhostUserBackend: Send + Sync {
     /// Get available vhost protocol features.
     fn protocol_features(&self) -> VhostUserProtocolFeatures;
 
+    /// Reset the emulated device state.
```
dorindabassey, line 62: small nitpick: I think you should also add documentation comment that this is a default implementation.

vhost-user-backend/src/backend.rs
```diff
@@ -177,6 +180,9 @@ pub trait VhostUserBackendMut: Send + Sync {
...
```
dorindabassey, line 183: same here

vhost-user-backend/src/backend.rs
XanClic: Sure!

XanClic:
v7: Add the ubiquitous “A default implementation is provided as we cannot expect all backends to implement this function.” note to the `VhostUserBackend{,Mut}::reset_device()` documentation

(Thanks for all the reviews so far :slightly_smiling_face:)

stefano-garzarella (approved):

Ablu (approved):

stefano-garzarella:
I'm thinking of delaying a bit merging new stuff, just to simplify the release with #267 if we need a vhost-user-backend-v0.16.1 tag.

dorindabassey (approved):

PR #267, opened 2024-10-18
Fix feature unification issues with gpu-socket feature

dorindabassey:
### Summary of the PR

make `set_gpu_socket` method a default implementation and remove conditional compilation because cargo feature unification already defeats the purpose of the `gpu-socket` feature flag.

Fixes #265 

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR have Signed-Off-By trailers (with
  `git commit -s`), and the commit message has max 60 characters for the
  summary and max 75 characters for each description line.
- [x] All added/changed functionality has a corresponding unit/integration
  test.
- [x] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [x] Any newly added `unsafe` code is properly documented.

PR #268, opened 2024-10-21
Add VHOST_USER_SHARED_OBJECT support

dorindabassey:
### Summary of the PR

This PR adds support for the VHOST_USER_GET_SHARED_OBJECT message as described in the QEMU [specification](https://qemu-project.gitlab.io/qemu/interop/vhost-user.html) 
When the VHOST_USER_GET_SHARED_OBJECT is triggered from a vhost device, this message is expected to return a fd with reference to the resources specified by the uuid. This message can be used if a vhost device export resources (SHARED_OBJECT_ADD), and the UUID for that resource is found in the exporters cache.  This message allows forwarding of the fd to another vhost device in form of SHARED_OBJECT_LOOKUP message which imports the resources.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-user-backend/Cargo.toml
+++ b/vhost-user-backend/Cargo.toml
@@ -24,6 +24,7 @@ vmm-sys-util = { workspace = true }
 
 [dev-dependencies]
 nix = { version = "0.29", features = ["fs"] }
+uuid = { version = "1.8.0", features=["v4"] }
 vhost = { path = "../vhost", version = "0.13.0", features = ["test-utils", "vhost-user-frontend", "vhost-user-backend"] }
 vm-memory = { workspace = true, features = ["backend-mmap", "backend-atomic"] }
 tempfile = "3.2.0"
--- a/vhost-user-backend/src/backend.rs
+++ b/vhost-user-backend/src/backend.rs
@@ -25,6 +25,7 @@ use std::sync::{Arc, Mutex, RwLock};
 
 use vhost::vhost_user::message::{
     VhostTransferStateDirection, VhostTransferStatePhase, VhostUserProtocolFeatures,
+    VhostUserSharedMsg,
 };
 use vhost::vhost_user::Backend;
 use vm_memory::bitmap::Bitmap;
@@ -91,6 +92,20 @@ pub trait VhostUserBackend: Send + Sync {
     /// A default implementation is provided as we cannot expect all backends to implement this
     /// function.
     fn set_backend_req_fd(&self, _backend: Backend) {}
+
+    /// This method retrieves a file descriptor for a shared object, identified by a unique UUID,
+    /// which can be used by the front-end for DMA. If the shared object is found, it must return
+    /// a File that the frontend can use. If the shared object does not exist the function returns
+    /// `None` (indicating no file descriptor is available).
+    ///
+    /// This function returns a `Result`, returning an error if the backend does not implement this
+    /// function.
+    fn get_shared_object(&self, _uuid: VhostUserSharedMsg) -> Result<File> {
+        Err(std::io::Error::new(
+            std::io::ErrorKind::Unsupported,
+            "back end does not support get shared object",
+        ))
+    }
 
     /// Set handler for communicating with the frontend by the gpu specific backend communication
     /// channel.
@@ -220,6 +235,20 @@ pub trait VhostUserBackendMut: Send + Sync {
     /// function.
     fn set_backend_req_fd(&mut self, _backend: Backend) {}
 
+    /// This method retrieves a file descriptor for a shared object, identified by a unique UUID,
+    /// which can be used by the front-end for DMA. If the shared object is found, it must return
+    /// a File that the frontend can use. If the shared object does not exist the function returns
+    /// `None` (indicating no file descriptor is available).
+    ///
+    /// This function returns a `Result`, returning an error if the backend does not implement this
+    /// function.
+    fn get_shared_object(&mut self, _uuid: VhostUserSharedMsg) -> Result<File> {
+        Err(std::io::Error::new(
+            std::io::ErrorKind::Unsupported,
+            "back end does not support get shared object",
+        ))
+    }
+
     /// Set handler for communicating with the frontend by the gpu specific backend communication
     /// channel.
     ///
@@ -341,6 +370,10 @@ impl<T: VhostUserBackend> VhostUserBackend for Arc<T> {
         self.deref().set_backend_req_fd(backend)
     }
 
+    fn get_shared_object(&self, uuid: VhostUserSharedMsg) -> Result<File> {
+        self.deref().get_shared_object(uuid)
+    }
+
     fn set_gpu_socket(&self, gpu_backend: GpuBackend) -> Result<()> {
         self.deref().set_gpu_socket(gpu_backend)
     }
@@ -424,6 +457,10 @@ impl<T: VhostUserBackendMut> VhostUserBackend for Mutex<T> {
 
     fn set_backend_req_fd(&self, backend: Backend) {
         self.lock().unwrap().set_backend_req_fd(backend)
+    }
+
+    fn get_shared_object(&self, uuid: VhostUserSharedMsg) -> Result<File> {
+        self.lock().unwrap().get_shared_object(uuid)
     }
 
     fn set_gpu_socket(&self, gpu_backend: GpuBackend) -> Result<()> {
@@ -512,6 +549,10 @@ impl<T: VhostUserBackendMut> VhostUserBackend for RwLock<T> {
 
     fn set_backend_req_fd(&self, backend: Backend) {
         self.write().unwrap().set_backend_req_fd(backend)
+    }
+
+    fn get_shared_object(&self, uuid: VhostUserSharedMsg) -> Result<File> {
+        self.write().unwrap().get_shared_object(uuid)
     }
 
     fn set_gpu_socket(&self, gpu_backend: GpuBackend) -> Result<()> {
--- a/vhost-user-backend/src/backend.rs
+++ b/vhost-user-backend/src/backend.rs
@@ -601,6 +601,7 @@ pub mod tests {
     use crate::VringRwLock;
     use libc::EFD_NONBLOCK;
     use std::sync::Mutex;
+    use uuid::Uuid;
     use vm_memory::{GuestAddress, GuestMemoryAtomic, GuestMemoryMmap};
 
     pub struct MockVhostBackend {
@@ -685,6 +686,11 @@ pub mod tests {
 
         fn set_backend_req_fd(&mut self, _backend: Backend) {}
 
+        fn get_shared_object(&mut self, _uuid: VhostUserSharedMsg) -> Result<File> {
+            let file = tempfile::tempfile().unwrap();
+            Ok(file)
+        }
+
         fn queues_per_thread(&self) -> Vec<u64> {
             vec![1, 1]
         }
@@ -735,6 +741,11 @@ pub mod tests {
 
         let _ = backend.exit_event(0).unwrap();
 
+        let uuid = VhostUserSharedMsg {
+            uuid: Uuid::new_v4(),
+        };
+        backend.get_shared_object(uuid).unwrap();
+
         let mem = GuestMemoryAtomic::new(
             GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0x100000), 0x10000)]).unwrap(),
         );
@@ -774,6 +785,11 @@ pub mod tests {
             GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0x100000), 0x10000)]).unwrap(),
         );
         backend.update_memory(mem.clone()).unwrap();
+
+        let uuid = VhostUserSharedMsg {
+            uuid: Uuid::new_v4(),
+        };
+        backend.get_shared_object(uuid).unwrap();
 
         let vring = VringRwLock::new(mem, 0x1000).unwrap();
         backend
--- a/vhost-user-backend/src/handler.rs
+++ b/vhost-user-backend/src/handler.rs
@@ -18,8 +18,9 @@ use crate::bitmap::{BitmapReplace, MemRegionBitmap, MmapLogReg};
 use userfaultfd::{Uffd, UffdBuilder};
 use vhost::vhost_user::message::{
     VhostTransferStateDirection, VhostTransferStatePhase, VhostUserConfigFlags, VhostUserLog,
-    VhostUserMemoryRegion, VhostUserProtocolFeatures, VhostUserSingleMemoryRegion,
-    VhostUserVirtioFeatures, VhostUserVringAddrFlags, VhostUserVringState,
+    VhostUserMemoryRegion, VhostUserProtocolFeatures, VhostUserSharedMsg,
+    VhostUserSingleMemoryRegion, VhostUserVirtioFeatures, VhostUserVringAddrFlags,
+    VhostUserVringState,
 };
 use vhost::vhost_user::GpuBackend;
 use vhost::vhost_user::{
@@ -572,6 +573,16 @@ where
         self.backend
             .set_gpu_socket(gpu_backend)
             .map_err(VhostUserError::ReqHandlerError)
+    }
+
+    fn get_shared_object(&mut self, uuid: VhostUserSharedMsg) -> VhostUserResult<File> {
+        match self.backend.get_shared_object(uuid) {
+            Ok(shared_file) => Ok(shared_file),
+            Err(e) => Err(VhostUserError::ReqHandlerError(io::Error::new(
+                io::ErrorKind::Other,
+                e,
+            ))),
+        }
     }
 
     fn get_inflight_fd(
--- a/vhost/src/vhost_user/backend_req_handler.rs
+++ b/vhost/src/vhost_user/backend_req_handler.rs
@@ -68,6 +68,7 @@ pub trait VhostUserBackendReqHandler {
     fn set_config(&self, offset: u32, buf: &[u8], flags: VhostUserConfigFlags) -> Result<()>;
     fn set_backend_req_fd(&self, _backend: Backend) {}
     fn set_gpu_socket(&self, _gpu_backend: GpuBackend) -> Result<()>;
+    fn get_shared_object(&self, uuid: VhostUserSharedMsg) -> Result<File>;
     fn get_inflight_fd(&self, inflight: &VhostUserInflight) -> Result<(VhostUserInflight, File)>;
     fn set_inflight_fd(&self, inflight: &VhostUserInflight, file: File) -> Result<()>;
     fn get_max_mem_slots(&self) -> Result<u64>;
@@ -129,6 +130,7 @@ pub trait VhostUserBackendReqHandlerMut {
     fn set_config(&mut self, offset: u32, buf: &[u8], flags: VhostUserConfigFlags) -> Result<()>;
     fn set_backend_req_fd(&mut self, _backend: Backend) {}
     fn set_gpu_socket(&mut self, _gpu_backend: GpuBackend) -> Result<()>;
+    fn get_shared_object(&mut self, uuid: VhostUserSharedMsg) -> Result<File>;
     fn get_inflight_fd(
         &mut self,
         inflight: &VhostUserInflight,
@@ -246,6 +248,10 @@ impl<T: VhostUserBackendReqHandlerMut> VhostUserBackendReqHandler for Mutex<T> {
 
     fn set_gpu_socket(&self, gpu_backend: GpuBackend) -> Result<()> {
         self.lock().unwrap().set_gpu_socket(gpu_backend)
+    }
+
+    fn get_shared_object(&self, uuid: VhostUserSharedMsg) -> Result<File> {
+        self.lock().unwrap().get_shared_object(uuid)
     }
 
     fn get_inflight_fd(&self, inflight: &VhostUserInflight) -> Result<(VhostUserInflight, File)> {
@@ -562,6 +568,26 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
                 self.check_request_size(&hdr, size, hdr.get_size() as usize)?;
                 let res = self.set_backend_req_fd(files);
                 self.send_ack_message(&hdr, res)?;
+            }
+            Ok(FrontendReq::GET_SHARED_OBJECT) => {
+                self.check_proto_feature(VhostUserProtocolFeatures::SHARED_OBJECT)?;
+                self.check_request_size(&hdr, size, hdr.get_size() as usize)?;
+                let msg = self.extract_request_body::<VhostUserSharedMsg>(&hdr, size, &buf)?;
+
+                let reply_hdr = self.new_reply_header::<VhostUserEmpty>(&hdr, 0)?;
+                match self.backend.get_shared_object(msg) {
+                    Ok(file) => {
+                        self.main_sock.send_message(
+                            &reply_hdr,
+                            &VhostUserEmpty,
+                            Some(&[file.as_raw_fd()]),
+                        )?;
+                    }
+                    Err(_) => {
+                        self.main_sock
+                            .send_message(&reply_hdr, &VhostUserEmpty, None)?;
+                    }
+                }
             }
             Ok(FrontendReq::GET_INFLIGHT_FD) => {
                 self.check_proto_feature(VhostUserProtocolFeatures::INFLIGHT_SHMFD)?;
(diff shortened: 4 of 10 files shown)


vhost-user-backend/src/backend.rs
```diff
@@ -87,6 +89,15 @@ pub trait VhostUserBackend: Send + Sync {
     /// function.
     fn set_backend_req_fd(&self, _backend: Backend) {}
 
+    /// This function gets an owned file descriptor that the front-end can use otherwise
+    /// no file descriptor.
+    ///
+    /// A default implementation is provided as we cannot expect all backends to implement this
+    /// function.
+    fn get_shared_object(&self, _uuid: VhostUserSharedMsg) -> Result<Option<OwnedFd>> {
+        Ok(None)
```
aesteve-rh, line 98: It is nice to have a default implementation, but I wonder why this differs from the `VhostUserBackendMut` counterpart implementation. IMO this could also fail with `Unsupported` error.

dorindabassey: yes, this was an oversight. thanks for pointing it out.

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -555,6 +561,24 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
...
+
+                match res {
+                    Ok(file) => {
+                        let hdr = self.new_reply_header::<VhostUserSharedMsg>(&hdr, 0)?;
+                        self.main_sock
+                            .send_message(&hdr, &msg, Some(&[file.as_raw_fd()]))?
+                    }
+                    Err(_) => {
+                        let msg = VhostUserSharedMsg::default();
+                        self.main_sock.send_message(&hdr, &msg, None)?;
```
stefano-garzarella, line 578: Why are we re-using the `hdr` of the request?
BTW also in this case I'm not sure if it's a good idea using `VhostUserSharedMsg`, so maybe just call   `self.send_ack_message(&hdr, res)?;` or send an empty message without any fd should be fine, the spec are not really clear.

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 78.40,
+  "coverage_score": 77.80,
```
stefano-garzarella, line 2: Please mention the drop in the commit description or try to add more tests.

vhost/src/vhost_user/dummy_backend.rs
```diff
@@ -262,6 +263,17 @@ impl VhostUserBackendReqHandlerMut for DummyBackendReqHandler {
     #[cfg(feature = "gpu-socket")]
     fn set_gpu_socket(&mut self, _gpu_backend: GpuBackend) {}
 
+    fn get_shared_object(&mut self, _uuid: VhostUserSharedMsg) -> Result<OwnedFd> {
+        let file = tempfile::tempfile().unwrap();
+
+        // SAFETY: We are calling `OwnedFd::from_raw_fd` to take ownership of the raw file descriptor.
+        // This is safe in this context because:
+        // 1. The file descriptor returned by `into_raw_fd()` is valid and open.
```
stefano-garzarella, line 271: But are we sure it remains open after the end of this block?

For example in `get_inflight_fd()` we stored it in `self.inflight_file`, so IMO we should do something similar, otherwise the file will be closed and the `fd` will be invalid just after the call of `get_shared_object`.

vhost/src/vhost_user/backend_req_handler.rs
dorindabassey: yes, sending an empty message is fine, also I will remove the `hdr` and `msg` of `VhostUserSharedMsg`.  Thanks :) 

> the spec are not really clear.

+1

vhost/src/vhost_user/dummy_backend.rs
stefano-garzarella: right, I missed the `into_raw_fd()` call. In that case, it consumes it, so the OwnedFd will take the ownership!

vhost/src/vhost_user/backend_req_handler.rs
aesteve-rh: Frontend implementation basically only checks for request type and the fd for the replies, so backends can do anything as long as they leave the ancillary fds empty. 

I think empty message is fine.

On the other hand, I think this is the first time someone looked into an actual implementation of the backend side to support this message type. If you think it would be more useful to define the return type otherwise, we could propose a change to the spec.

stefano-garzarella: IIUC most commands return a `u64` with some kind of return value, zero on success or non-zero on failure. There is also the `VHOST_USER_PROTOCOL_F_REPLY_ACK` section talking about that. So maybe we should follow that, but IMHO that should be more clear in the spec in general, not for this specific command.

vhost-user-backend/src/backend.rs
```diff
@@ -87,6 +89,18 @@ pub trait VhostUserBackend: Send + Sync {
     /// function.
     fn set_backend_req_fd(&self, _backend: Backend) {}
 
+    /// This function gets an owned file descriptor that the front-end can use otherwise
+    /// no file descriptor.
+    ///
+    /// A default implementation is provided as we cannot expect all backends to implement this
```
stefano-garzarella, line 95: I'd update this comment like we did for PR #267, to make clear that backends need to implement it if the feature is supported, otherwise we just return an error.

vhost-user-backend/src/backend.rs
```diff
@@ -87,6 +89,18 @@ pub trait VhostUserBackend: Send + Sync {
     /// function.
     fn set_backend_req_fd(&self, _backend: Backend) {}
 
+    /// This function gets an owned file descriptor that the front-end can use otherwise
```
stefano-garzarella, line 92: `the front-end can use otherwise no file descritptor`, What would you like to say here? It sounds like a cut sentence.

vhost-user-backend/src/backend.rs
```diff
@@ -205,6 +219,18 @@ pub trait VhostUserBackendMut: Send + Sync {
     /// function.
     fn set_backend_req_fd(&mut self, _backend: Backend) {}
 
+    /// This function gets an owned file descriptor that the front-end can use otherwise
+    /// no file descriptor.
+    ///
+    /// A default implementation is provided as we cannot expect all backends to implement this
+    /// function.
```
stefano-garzarella, line 226: Ditto for both (see previous comments)

vhost-user-backend/Cargo.toml
```diff
@@ -17,6 +17,7 @@ gpu-socket = ["vhost/gpu-socket"]
 libc = "0.2.39"
 log = "0.4.17"
 userfaultfd = { version = "0.8.1", optional = true }
+uuid = { version = "1.8.0", features=["v4", "fast-rng", "macro-diagnostics"] }
```
stefano-garzarella, line 20: Can you explain why we need these 3 features enabled in the commit description?

stefano-garzarella: IIUC this is used just for testing, so it should go in `dev-dependencies` section if it's the case.

stefano-garzarella (changes requested):

vhost/src/vhost_user/backend_req_handler.rs
stefano-garzarella: The issue with `hdr` is still there. Can you define a `reply_hdr` just before the match, to be sure that you're not reusing the incoming `hdr` ?

vhost/src/vhost_user/dummy_backend.rs
```diff
@@ -262,6 +263,17 @@ impl VhostUserBackendReqHandlerMut for DummyBackendReqHandler {
...
+        // 2. We ensure that `OwnedFd` will properly close the file descriptor when it is dropped.
+        let owned_fd = unsafe { OwnedFd::from_raw_fd(file.into_raw_fd()) };
+        Ok(owned_fd)
```
stefano-garzarella, line 274: Sorry, but this code is still obscure to me, do you have an implementation where this is tested?

My main question is: Why do we return an OwnedFd?
IMO we should do it like for inflight_fd, because I expect a backend to open a file, and then keep it open. When a frontend ask for it, then clone the file or fd and pass it to the caller of `get_shared_object()`, but it should not give ownership to the caller, but just a clone, because otherwise the caller closes it just after sending the message to the frontend, or did I miss something?

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -555,6 +561,26 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
...
+                    Ok(file) => {
+                        let hdr = self.new_reply_header::<VhostUserEmpty>(&hdr, 0)?;
+                        self.main_sock.send_message(
+                            &hdr,
+                            &VhostUserEmpty,
+                            Some(&[file.as_raw_fd()]),
+                        )?;
+                    }
+                    Err(_) => {
+                        self.main_sock.send_message(&hdr, &VhostUserEmpty, None)?;
```
stefano-garzarella, line 580: `hdr` here is still the old one.

vhost/src/vhost_user/dummy_backend.rs
dorindabassey: this is supposed to be tested on another vhost-device, no implementation in place yet, but the vhost-device-gpu has this resource_assign_uuid command type that exports a resource uuid to a table using the `shared_object_add`  implementation (for reference: [draft](https://github.com/dorindabassey/vhost-device/pull/36) needs this qemu [patch](https://mail.gnu.org/archive/html/qemu-devel/2024-10/msg02865.html)) with the expectation that another vhost device will import a shared object using the `shared_object_lookup`. no implementation yet as this is meant to be done on another vhost-device. but testing locally is by calling the  `shared_object_lookup` which triggers QEMU to use the `VHOST_USER_GET_SHARED_OBJECT` command.

> My main question is: Why do we return an OwnedFd?

` OwnedFd` is more preferred for resources that use file descriptors but are not necessarily files, see  
[252 comment](https://github.com/rust-vmm/vhost/pull/252#issuecomment-2321315593)

> IMO we should do it like for inflight_fd, because I expect a backend to open a file, and then keep it open. When a frontend ask for it, then clone the file or fd and pass it to the caller of get_shared_object(), but it should not give ownership to the caller, but just a clone, because otherwise the caller closes it just after sending the message to the frontend, or did I miss something?

IIUC, your assumption is that immediately after the call to `get_shared_object()` the fd is closed, but what happens in this case is that it transfers ownership of the file descriptor to the caller of `get_shared_object()` which use the `OwnedFd` within the scope, and the descriptor will automatically close once it goes out of scope, not immediately after the call to `get_shared_object()`.
tbh, I have no strong opposition to using `File` though and i can change it to return `File`

vhost-user-backend/Cargo.toml
dorindabassey: > Can you explain why we need these 3 features enabled in the commit description?

we need just one `v4`, thanks for pointing that out, I will remove the other features.

> IIUC this is used just for testing, so it should go in dev-dependencies section if it's the case.

Ack

vhost/src/vhost_user/dummy_backend.rs
stefano-garzarella: > > My main question is: Why do we return an OwnedFd?
> 
> ` OwnedFd` is more preferred for resources that use file descriptors but are not necessarily files, see [252 comment](https://github.com/rust-vmm/vhost/pull/252#issuecomment-2321315593)

Okay, but why OwnedFd and not BorrowedFd ?
I mean my question was why passing the ownership of the fd to the caller?

> 
> > IMO we should do it like for inflight_fd, because I expect a backend to open a file, and then keep it open. When a frontend ask for it, then clone the file or fd and pass it to the caller of get_shared_object(), but it should not give ownership to the caller, but just a clone, because otherwise the caller closes it just after sending the message to the frontend, or did I miss something?
> 
> IIUC, your assumption is that immediately after the call to `get_shared_object()` the fd is closed, but what happens in this case is that it transfers ownership of the file descriptor to the caller of `get_shared_object()` which use the `OwnedFd` within the scope, and the descriptor will automatically close once it goes out of scope, not immediately after the call to `get_shared_object()`. tbh, I have no strong opposition to using `File` though and i can change it to return `File`

Just to be clear, BorrowedFd is fine by me, but I still don't get why OwnedFd.

Can you give an example of a backend implementing it?
For example, for me this example in a dummy backend is not clear.
Are we expecting a backend to open that fd just when  `get_shared_object()` is called, then forget about that object?
Maybe I'm missing something, since I don't know how dmabuf works, so a better commit description could help.

aesteve-rh: In principle, the importer will get the fd and attach itself to the buffer, but the exporter shall retain ownership. More than one importer is allowed, with the refcount increased for each one.

So I agree that it shouldn't be `OwnedFd`.

dorindabassey: > Just to be clear, BorrowedFd is fine by me, but I still don't get why OwnedFd.

Ok, BorrowedFd it is.

> In principle, the importer will get the fd and attach itself to the buffer, but the exporter shall retain ownership. More than one importer is allowed, with the refcount increased for each one.

> So I agree that it shouldn't be OwnedFd.

Noted, i saw that in [virtio-dmabuf](https://gitlab.com/qemu-project/qemu/-/blame/master/include/hw/virtio/virtio-dmabuf.h?ref_type=heads#L50-L51) and assumed that both importer and exporter needed ownership.

> Can you give an example of a backend implementing it?
For example, for me this example in a dummy backend is not clear.
Are we expecting a backend to open that fd just when get_shared_object() is called, then forget about that object?

currently, no backend is implementing this, but the next plan is to implement it in the vhost-device-gpu which will return the dmabuf fd associated with the uuid to the `importer that will get the fd and attach itself to the buffer` when a lookup message is sent from another vhost device.
i don't mind using borrowed fd that way we keep a temporary access to the fd without transferring ownership.

vhost/src/vhost_user/backend_req_handler.rs
dorindabassey: Thanks, will fix it. I copied what postcopy_advice() did, I think I should change that aswell in another PR.

vhost/src/vhost_user/dummy_backend.rs
aesteve-rh: > Noted, i saw that in [virtio-dmabuf](https://gitlab.com/qemu-project/qemu/-/blame/master/include/hw/virtio/virtio-dmabuf.h?ref_type=heads#L50-L51) and assumed that both importer and exporter needed ownership.

Uhm, in the linked doc it states that caller retains ownership, being the caller, the vhost-user handling code. It is a developer's detail. Just because the virtio-dmabuf API is storing a pointer, but has no control over its life cycle.

stefano-garzarella: > currently, no backend is implementing this, but the next plan is to implement it in the vhost-device-gpu which will return the dmabuf fd associated with the uuid to the `importer that will get the fd and attach itself to the buffer` when a lookup message is sent from another vhost device. 

My suggestion for the future, when adding new API, is to provide also a real example. I guess with a real device, we can see that the fd's ownership can or not be passed to the caller. This will help and speed up the review phase IMHO.

> i don't mind using borrowed fd that way we keep a temporary access to the fd without transferring ownership.

Honestly, I can't see how it can work in a backend with a OwnedFd, for that reason I was asking for a working example. Also because I don't know anything about dmabuf, so I may be wrong on some assumption.

vhost/src/vhost_user/backend_req_handler.rs
stefano-garzarella: Yeah, I think so, and I agree on a separated PR, thanks in advance for that ;-)

vhost/src/vhost_user/dummy_backend.rs
dorindabassey: I just realized that returning `borrowedFd` implicitly means that i have to use `as_raw_fd()` to get the raw fd when creating a BorrowedFd. this also means that the trait implementation will error `cannot return value referencing temporary value` which is related to the lifetime of the borrowed value. Hence, in a bid to solve this I use `as_raw_fd()` in the backend_req_handler.rs to get the raw fd for the BorrowedFd. I mean this is what we are trying to avoid already - #252 + it's introducing a lot of unsafe blocks.  so I will just stick to using File, I believe that's a more simpler approach.

vhost-user-backend/src/handler.rs
```diff
@@ -574,6 +575,17 @@ where
             .map_err(VhostUserError::ReqHandlerError)
     }
 
+    fn get_shared_object(&mut self, uuid: VhostUserSharedMsg) -> VhostUserResult<File> {
+        match self.backend.get_shared_object(uuid) {
+            Ok(Some(shared_file)) => Ok(shared_file),
+            Ok(None) => Err(VhostUserError::IncorrectFds),
```
stefano-garzarella, line 581: We are mapping `None` on an VhostUserError, so can we just return `Result<File>` from `VhostUserBackend::get_shared_object` ?

vhost-user-backend/src/backend.rs
```diff
@@ -92,6 +93,20 @@ pub trait VhostUserBackend: Send + Sync {
...
     fn set_backend_req_fd(&self, _backend: Backend) {}
 
+    /// This method retrieves a file descriptor for a shared object, identified by a unique UUID,
+    /// which can be used by the front-end for DMA. If the shared object is found, it must return
+    /// a File that the frontend can use. If the shared object does not exist the function returns
+    /// `None` (indicating no file descriptor is available).
+    ///
+    /// This function returns a `Result`, returning an error if the backend does not implement this
+    /// function.
+    fn get_shared_object(&self, _uuid: VhostUserSharedMsg) -> Result<Option<File>> {
```
stefano-garzarella, line 103: ... I mean here returning `Result<File>`

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -563,6 +569,27 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
...
                 self.send_ack_message(&hdr, res)?;
             }
+            Ok(FrontendReq::GET_SHARED_OBJECT) => {
+                self.check_proto_feature(VhostUserProtocolFeatures::SHARED_OBJECT)?;
+                self.check_request_size(&hdr, size, hdr.get_size() as usize)?;
+                let msg = self.extract_request_body::<VhostUserSharedMsg>(&hdr, size, &buf)?;
+                let res = self.backend.get_shared_object(msg);
+                let reply_hdr = self.new_reply_header::<VhostUserEmpty>(&hdr, 0)?;
+
+                match res {
```
stefano-garzarella, line 579: `res` is difficult to understand and also not needed, so I'd suggest to remove it and call `get_shared_object` here:

```diff
diff --git a/vhost/src/vhost_user/backend_req_handler.rs b/vhost/src/vhost_user/backend_req_handler.rs
index 07e01a9..14290bb 100644
--- a/vhost/src/vhost_user/backend_req_handler.rs
+++ b/vhost/src/vhost_user/backend_req_handler.rs
@@ -573,10 +573,9 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
                 self.check_proto_feature(VhostUserProtocolFeatures::SHARED_OBJECT)?;
                 self.check_request_size(&hdr, size, hdr.get_size() as usize)?;
                 let msg = self.extract_request_body::<VhostUserSharedMsg>(&hdr, size, &buf)?;
-                let res = self.backend.get_shared_object(msg);
-                let reply_hdr = self.new_reply_header::<VhostUserEmpty>(&hdr, 0)?;
 
-                match res {
+                let reply_hdr = self.new_reply_header::<VhostUserEmpty>(&hdr, 0)?;
+                match self.backend.get_shared_object(msg) {
                     Ok(file) => {
                         self.main_sock.send_message(
                             &reply_hdr,
```

vhost/src/vhost_user/dummy_backend.rs
stefano-garzarella: We already use File everywhere, so I can't disagree with your change, but in the future I would like to switch to OwnedFd/BorrowedFd everywhere. For now I understand that it doesn't make sense to have API with File and API with BorrowedFd.

stefano-garzarella (commented):
The code LGTM, if you already have a real backend using this, I'd like to see that code to check that this API will work with it.

dorindabassey:
> The code LGTM, if you already have a real backend using this, I'd like to see that code to check that this API will work with it.

Hi Stefano, No real backend is using this yet, but I can perform a `shared-object lookup` request to frontend from the vhost_device_gpu which in turn triggers the `get_shared_object`. Ideally, though, a proper implementation should be setup from a virtio-video or virtio-media and not from the gpu device that adds and removes its object from the shared lookup table. I had tried to set that up with vhost-device-video, but the device is not fully matured to use the shared_object as it will require some more effort. In summary, there is no support for it at the moment.

stefano-garzarella:
@dorindabassey okay, so what about waiting till we have some backend using this?
I'm just concerned about introducing something that we then have to change, so I'd like to do that when we have code that uses it.

dorindabassey:
> @dorindabassey okay, so what about waiting till we have some backend using this? I'm just concerned about introducing something that we then have to change, so I'd like to do that when we have code that uses it.

In that case, this PR can wait, that way we can be sure it aligns with the requirement for the `get_shared_object`.
I actually thought a dummy backend would be sufficient for testing this.

stefano-garzarella:
> @dorindabassey okay, so what about waiting till we have some backend using this? I'm just concerned about introducing something that we then have to change, so I'd like to do that when we have code that uses it.

@aesteve-rh @slp any opinion on that?
I don't know shared object very well, but if you think the API is fine, we can just merge it and refine the implementation later.

aesteve-rh:
> @aesteve-rh @slp any opinion on that?
I don't know shared object very well, but if you think the API is fine, we can just merge it and refine the implementation later.

The API looks good to me. And I do not think this would break anything, as it completely relies on the backends implementing the `get_shared_object` method, and otherwise the feature remains unsupported even if a VMM send the request. So once we really start testing this with real devices, I think most of the work would be deferred to the backend implementation, and this part should remain mostly untouched.

That said, I agree it would be interesting to see how this looks like from the backend PoV. And I'd bet we are the first to test shared objects, so there is no rush to integrate. We need both exporter and importer to be able to test this properly.

So if you ask me, I think I'd go ahead and merge this. But I just wanted to highlight that I do understand the hesitation on integrating a partially untested code.

stefano-garzarella:
> So if you ask me, I think I'd go ahead and merge this. But I just wanted to highlight that I do understand the hesitation on integrating a partially untested code.

@aesteve-rh I see your point, so let's go in that direction. Can you review and approve this?

@dorindabassey I rebased and marked ready, let me know if we need to do something else.

stefano-garzarella (approved):

Ablu (commented):
I must admit that reading https://qemu-project.gitlab.io/qemu/interop/vhost-user.html has not helped me to figure out what this shared object feature is used for. But code looks OK to me apart from the minor comment.

aesteve-rh:
@Ablu The feature represents an implementation of the exporting shared objects, briefly described in the virtio specs: https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-10500011

I must admit that the explanation is not too detailed either. Essentially, it is a way for two virtio devices to share a dmabuf in the host. In the driver-side inside the guest kernel, the dmabuf is identified by the UUID.

So, in the host-side, it is required a table that relates UUIDs with the actual objects (i.e., the dmabufs). For vhost-user devices the table must be maintained by the devices themselves. The `get` message implemented in this PR is a way for the VMM to retrieve the object from the device, when requested by the importer device.

As of virtio spec v.1.3, there is no virtio device that would act as an importer (being virtio-gpu the unique exporter), and thus this feature still hasn't gotten a lot of attention.

vhost-user-backend/src/handler.rs
```diff
@@ -574,6 +575,17 @@ where
             .map_err(VhostUserError::ReqHandlerError)
     }
 
+    fn get_shared_object(&mut self, uuid: VhostUserSharedMsg) -> VhostUserResult<File> {
+        println!("vhost get shared object call");
```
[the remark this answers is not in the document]
dorindabassey: yes, I forgot that. I was using it for my debugging thanks!

stefano-garzarella (approved):

Ablu (approved):

stefano-garzarella (approved):

aesteve-rh (approved):

PR #271, opened 2024-10-25
vhost v0.12.1 and vhost-user-backend v0.16.1 release

stefano-garzarella:
### Summary of the PR

Remove the gpu-socket feature just introduced, since it doesn't work well with cargo feature unification.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-user-backend/CHANGELOG.md
+++ b/vhost-user-backend/CHANGELOG.md
@@ -8,6 +8,10 @@
 ### Deprecated
 
 ### Fixed
+
+## v0.16.1
+
+### Fixed
 - [[#267](https://github.com/rust-vmm/vhost/pull/267)] Fix feature unification issues with gpu-socket feature.
 
 ## v0.16.0
--- a/vhost-user-backend/Cargo.toml
+++ b/vhost-user-backend/Cargo.toml
@@ -16,7 +16,7 @@ postcopy = ["vhost/postcopy", "userfaultfd"]
 libc = "0.2.39"
 log = "0.4.17"
 userfaultfd = { version = "0.8.1", optional = true }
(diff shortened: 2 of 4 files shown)


stefano-garzarella:
After these releases, I'm undecided whether to yank the previous two. I don't expect any backend is using gpu_socket, so I'm not sure if it's necessary or not.

Opinions?

TimePrinciple:
> After these releases, I'm undecided whether to yank the previous two. I don't expect any backend is using gpu_socket, so I'm not sure if it's necessary or not.
> 
> Opinions?

IMHO if that's a serious problem, we should yank it to let downstream notice that change and fix it accordingly. I'm not an expert on that so I can only say it's up to you to decide how serious it is :)

stefano-garzarella:
> > After these releases, I'm undecided whether to yank the previous two. I don't expect any backend is using gpu_socket, so I'm not sure if it's necessary or not.
> > Opinions?
> 
> IMHO if that's a serious problem, we should yank it to let downstream notice that change and fix it accordingly. I'm not an expert on that so I can only say it's up to you to decide how serious it is :)

Yeah, and maybe also to avoid future backends using those versions. I thought it can affect just gpu backends, but we saw in vhost-device that the feature unification affects other, so yep, I'll yank it.

Ablu:
LGTM! But we should also document in the changelog why we yanked versions IMHO. But that can happen separately of course.

Ablu (approved):

rbradford (approved):

stefano-garzarella:
> LGTM! But we should also document in the changelog why we yanked versions IMHO. But that can happen separately of course.

Yep, good point. I'll open another PR with that changes and I'll yank them only after that PR will be merged!

PR #273, opened 2024-10-25
Yank vhost v0.12.0 and vhost-user-backend v0.16.0

stefano-garzarella:
### Summary of the PR

These versions got yanked because the `gpu_socket` feature introduced
in this release was causing problems (see [#265](https://github.com/rust-vmm/vhost/issues/265)).
Starting with the next versions, the `gpu_socket` feature was removed.

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR have Signed-Off-By trailers (with
  `git commit -s`), and the commit message has max 60 characters for the
  summary and max 75 characters for each description line.
- [x] All added/changed functionality has a corresponding unit/integration
  test.
- [x] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [x] Any newly added `unsafe` code is properly documented.

PR #274, opened 2024-10-25
CHANGELOGs: uniform and fix some inconsistencies

stefano-garzarella:
### Summary of the PR

Every time I do a release I notice this difference between
changelogs in this workspace, so let's try to uniform them by using
the form used in vhost-user-backend: without square brackets in the
title of the release and the `v` as prefix (v0.0.0).

Also fixed some difference in how we link PRs, let's follow the
most used way: [[#000]](url)

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-user-backend/CHANGELOG.md
+++ b/vhost-user-backend/CHANGELOG.md
@@ -20,7 +20,7 @@
 ## v0.16.1
 
 ### Fixed
-- [[#267](https://github.com/rust-vmm/vhost/pull/267)] Fix feature unification issues with gpu-socket feature.
+- [[#267]](https://github.com/rust-vmm/vhost/pull/267) Fix feature unification issues with gpu-socket feature.
 
 ## v0.16.0 - yanked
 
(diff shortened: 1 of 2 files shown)


stefano-garzarella:
I based this on top of #273 so we don't need to fix the conflicts. So marking as draft till we have it merged.

TimePrinciple:
Maybe we could use mdformat that kind of thing to take care of formatting on CHANGELOG.md, then people won't easily break consistency 🤔

stefano-garzarella:
> Maybe we could use mdformat that kind of thing to take care of formatting on CHANGELOG.md, then people won't easily break consistency 🤔

Do you mean in the CI?

TimePrinciple:
> > Maybe we could use mdformat that kind of thing to take care of formatting on CHANGELOG.md, then people won't easily break consistency 🤔
> 
> Do you mean in the CI?

Yes, we can put some checks (run some formatters) in our CI :)

PR #277, opened 2024-11-11
vhost: Refactor reply header initialization in POSTCOPY_ADVISE

dorindabassey:
### Summary of the PR

In BackendReqHandler::POSTCOPY_ADVISE, refactor the reply header
initialization for clarity and reuse.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost/CHANGELOG.md
+++ b/vhost/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Deprecated
 
 ### Fixed
+- [[#277]](https://github.com/rust-vmm/vhost/pull/277) vhost: Fix reply header for error case in POSTCOPY_ADVISE
 
 ## v0.13.0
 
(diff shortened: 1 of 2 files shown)


stefano-garzarella:
This is more of a fix than just a refactoring, so I would make that clear in the title and description.
Something like:
```
vhost: fix reply header for error case in POSTCOPY_ADVISE

...
```

Please add this PR in the changelog as well, since this fix the communication with the frontend.

dorindabassey:
> This is more of a fix than just a refactoring, so I would make that clear in the title and description. Something like:
> 
> ```
> vhost: fix reply header for error case in POSTCOPY_ADVISE
> 
> ...
> ```
> 
> Please add this PR in the changelog as well, since this fix the communication with the frontend.

Ok noted, although I didn't think it needed a changelog since it was quite small.

stefano-garzarella (approved):

Ablu (approved):

PR #284, opened 2025-03-08
Added function get_enabled to export vring enable status.

WeiChungHsu:
### Summary of the PR
Added function get_enabled to export vring enable status.

I am working on snapshot/restore feature for vhost-device-vsock crate. And during snapshot, system sends "FrontendReq::SET_VRING_ENABLE, enable false" to vhost-device-vsock device. Then the vring was disabled. However, the vring tx/rx queue still execute iter() which will return error and then system (vring) hang because vring has been disabled.

My proposed solution to solve this snapshot/restore hang issue is adding vring enable status checking before access/iter() the vring. Therefore, we need this get_enabled function to check if vring is enabled or disabled. I believe in the future, other vhost device may have similar hang issue when it executes snapshot.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost/src/lib.rs
+++ b/vhost/src/lib.rs
@@ -1,7 +1,7 @@
 // Copyright (C) 2019 Alibaba Cloud. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
 
-//! Virtio Vhost Backend Drivers
+//! # Virtio Vhost Backend Drivers
 //!
 //! Virtio devices use virtqueues to transport data efficiently. The first generation of virtqueue
 //! is a set of three different single-producer, single-consumer ring structures designed to store


Ablu:
Do you have an example how the corresponding change looks on the `vhost-device-vsock` side?

WeiChungHsu:
> Do you have an example how the corresponding change looks on the `vhost-device-vsock` side?

When the Vring is disabled (and queue set to not ready), the [vhu_vsock_thread.rs](http://vhu_vsock_thread.rs/) still try to iterate and read RX queue, therefore, it hang at this line:
https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-vsock/src/vhu_vsock_thread.rs#L591-L594

Furthermore, it hang because this Queue iter() operation return error at this line:
https://github.com/rust-vmm/vm-virtio/blob/main/virtio-queue/src/queue.rs#L592-L597

My solution in vhu_vsock_thread.rs:
        let mut vring_mut = vring.get_mut();

	if !vring_mut.get_enabled() {
	    return Ok(true);
	}

        let queue = vring_mut.get_queue_mut();

        while let Some(mut avail_desc) = queue
            .iter(atomic_mem.memory())
            .map_err(|_| Error::IterateQueue)?
            .next()

Ablu:
> Furthermore, it hang because this Queue iter() operation return error at this line:
> https://github.com/rust-vmm/vm-virtio/blob/main/virtio-queue/src/queue.rs#L592-L597

Hm. But shouldn't it error out in that case versus hanging? I do not fully understand where the hang would happen.

Ideally, I think it would be great if we can avoid easy footguns by designing the API in a way that avoids them altogether. I assume chances are that people will forget this check in other implementations too?

stefano-garzarella:
@WeiChungHsu can you fix CI issues and rebase this?

WeiChungHsu:
> Do you have an example how the corresponding change looks on the `vhost-device-vsock` side?

Hi,

Couple days ago, I created a PR for `vhost-device-vsock` side. It used this new vring function to check if vring is enabled/disabled.
https://github.com/rust-vmm/vhost-device/pull/827

Sorry for my late reply.

WeiChungHsu:
> @WeiChungHsu can you fix CI issues and rebase this?

Yes, I am working on it now.

stefano-garzarella:
@WeiChungHsu please avoid merging from main, instead rebase this PR on current main.

PR #291, opened 2025-03-28
Added function get_enabled to export vring enable status.

WeiChungHsu:
### Summary of the PR

The vring is disabled and RX/TX queue is set to
not-ready during snapshot. But vhost-device process
keeps access vring and queue after snapshot which
caused system hang. We need to export vring enabled
and disabled status for snapshot feature so that
the vhost-device would skip vring/queue access
after snapshot to avoid hang.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-user-backend/src/vring.rs
+++ b/vhost-user-backend/src/vring.rs
@@ -61,6 +61,9 @@ pub trait VringT<M: GuestAddressSpace>:
     /// Set vring enabled state.
     fn set_enabled(&self, enabled: bool);
 
+    /// Get vring enabled state.
+    fn get_enabled(&self) -> bool;
+
     /// Set queue addresses for descriptor table, available ring and used ring.
     fn set_queue_info(
         &self,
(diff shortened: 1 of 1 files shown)


WeiChungHsu:
> @WeiChungHsu why closing #284 in favor of this? What are the differences?
> 
> Please in the future avoid that, it's better to continue on the same PR to check the history of reviews, etc.

Sorry, I tried many git commands, but can't rebase this PR on current main. In addition, I accidentally close the old PR. Sorry about that.

stefano-garzarella:
There are questions still open like: https://github.com/rust-vmm/vhost/pull/284#issuecomment-2712859133

> >   Furthermore, it hang because this Queue iter() operation return error at this line:
> >   https://github.com/rust-vmm/vm-virtio/blob/main/virtio-queue/src/queue.rs#L592-L597

> Hm. But shouldn't it error out in that case versus hanging? I do not fully understand where the hang would happen.
> 
> Ideally, I think it would be great if we can avoid easy footguns by designing the API in a way that avoids them altogether. > I assume chances are that people will forget this check in other implementations too?

vhost-user-backend/src/vring.rs
```diff
@@ -61,6 +61,9 @@ pub trait VringT<M: GuestAddressSpace>:
     /// Set vring enabled state.
     fn set_enabled(&self, enabled: bool);
 
+    /// Get vring enabled state.
+    fn get_enabled(&self) -> bool;
```
bilelmoussaoui, line 65: This should be named is_enabled.

germag:
@WeiChungHsu any news on this?

PR #308, opened 2025-07-09
Replace Eventfd with EventNotifier/EventConsumer

uran0sH:
### Summary of the PR

Eventfd is Linux-specific. To support more platforms, we replace it with
the EventNotifier/EventConsumer abstractions.
EventSender and EventReceiver are wrappers that encapsulate eventfd functionality
Use pipefd to replace eventfd in the test.

### Requirements

requirements:

- [ x ] All commits in this PR have Signed-Off-By trailers (with
- [ x ] All added/changed functionality has a corresponding unit/integration
  test.
- [ x ] All added/changed public-facing functionality has entries in the "Upcoming 
- [ x ] Any newly added `unsafe` code is properly documented.

### DIFF
--- a/vhost-user-backend/src/backend.rs
+++ b/vhost-user-backend/src/backend.rs
@@ -30,7 +30,7 @@ use vhost::vhost_user::message::{
 use vhost::vhost_user::Backend;
 use vm_memory::bitmap::Bitmap;
 use vmm_sys_util::epoll::EventSet;
-use vmm_sys_util::eventfd::EventFd;
+use vmm_sys_util::event::{EventConsumer, EventNotifier};
 
 use vhost::vhost_user::GpuBackend;
 
@@ -132,7 +132,8 @@ pub trait VhostUserBackend: Send + Sync {
     ///
     /// The returned `EventFd` will be monitored for IO events. When the
     /// returned EventFd is written to, the worker thread will exit.
-    fn exit_event(&self, _thread_index: usize) -> Option<EventFd> {
+    // TODO: Refine this API to return only EventNotifier.
+    fn exit_event(&self, _thread_index: usize) -> Option<(EventConsumer, EventNotifier)> {
         None
     }
 
@@ -275,7 +276,8 @@ pub trait VhostUserBackendMut: Send + Sync {
     /// If an (`EventFd`, `token`) pair is returned, the returned `EventFd` will be monitored for IO
     /// events by using epoll with the specified `token`. When the returned EventFd is written to,
     /// the worker thread will exit.
-    fn exit_event(&self, _thread_index: usize) -> Option<EventFd> {
+    // TODO: Refine this API to return only EventNotifier.
+    fn exit_event(&self, _thread_index: usize) -> Option<(EventConsumer, EventNotifier)> {
         None
     }
 
@@ -382,7 +384,7 @@ impl<T: VhostUserBackend> VhostUserBackend for Arc<T> {
         self.deref().queues_per_thread()
     }
 
-    fn exit_event(&self, thread_index: usize) -> Option<EventFd> {
+    fn exit_event(&self, thread_index: usize) -> Option<(EventConsumer, EventNotifier)> {
         self.deref().exit_event(thread_index)
     }
 
@@ -471,7 +473,7 @@ impl<T: VhostUserBackendMut> VhostUserBackend for Mutex<T> {
         self.lock().unwrap().queues_per_thread()
     }
 
-    fn exit_event(&self, thread_index: usize) -> Option<EventFd> {
+    fn exit_event(&self, thread_index: usize) -> Option<(EventConsumer, EventNotifier)> {
         self.lock().unwrap().exit_event(thread_index)
     }
 
@@ -563,7 +565,7 @@ impl<T: VhostUserBackendMut> VhostUserBackend for RwLock<T> {
         self.read().unwrap().queues_per_thread()
     }
 
-    fn exit_event(&self, thread_index: usize) -> Option<EventFd> {
+    fn exit_event(&self, thread_index: usize) -> Option<(EventConsumer, EventNotifier)> {
         self.read().unwrap().exit_event(thread_index)
     }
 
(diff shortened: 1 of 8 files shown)


uran0sH:
cc@stefano-garzarella @germag @osteffenrh

vhost-user-backend/src/backend.rs
```diff
@@ -624,7 +624,21 @@ pub mod tests {
             // order to allow tests maximum flexibility in checking whether
             // signals arrived or not.
             backend.exit_event_fds = (0..backend.queues_per_thread().len())
-                .map(|_| EventFd::new(EFD_NONBLOCK).unwrap())
+                .map(|_| unsafe {
+                    let (reader, writer) = std::io::pipe().expect("Failed to create pipe");
```
stefano-garzarella: It's just a mock, but maybe this is a cool use case of what @roypat suggested in https://github.com/rust-vmm/vmm-sys-util/pull/244#pullrequestreview-3002187055

stefano-garzarella:
@uran0sH marked as draft, since we need a new vmm-sys-util release before merging this.

stefano-garzarella:
@uran0sH please rebase and check CI (come failures need the new rust version we are updating in the CI, but others are related to format, etc.)

vhost-user-backend/src/backend.rs
```diff
@@ -132,7 +132,7 @@ pub trait VhostUserBackend: Send + Sync {
     ///
     /// The returned `EventFd` will be monitored for IO events. When the
     /// returned EventFd is written to, the worker thread will exit.
-    fn exit_event(&self, _thread_index: usize) -> Option<EventFd> {
+    fn exit_event(&self, _thread_index: usize) -> Option<(EventConsumer, EventNotifier)> {
```
germag: Last time we discussed that this is fine because we thought it's an internal method, but It's public because it is part of a public trait: `pub trait VhostUserBackend` (We override it in virtiofsd, for instance)

vhost-user-backend/src/backend.rs
```diff
@@ -275,7 +275,7 @@ pub trait VhostUserBackendMut: Send + Sync {
     /// If an (`EventFd`, `token`) pair is returned, the returned `EventFd` will be monitored for IO
     /// events by using epoll with the specified `token`. When the returned EventFd is written to,
     /// the worker thread will exit.
-    fn exit_event(&self, _thread_index: usize) -> Option<EventFd> {
+    fn exit_event(&self, _thread_index: usize) -> Option<(EventConsumer, EventNotifier)> {
```
germag: ditto

vhost-user-backend/src/backend.rs
stefano-garzarella: Oh I see, so IMO you were right, this should only return `EventNotifier`, we may need to add another internal API to be used to get `EventConsumer` or find another way

bobosco: https://github.com/rust-vmm/vhost/blob/main/vhost-user-backend/src/event_loop.rs#L95 this line should need an EventConsumer, right? I have no idea how to solve it if we only return EventNotifer. And I don't exactly understand why it must only return EventNotifier but not both.

stefano-garzarella: Okay, I checked it and it's really a strange API design IMO (pre-existing).
It seems that every backend in rust-vmm/vhost-device override it with an EventFD and then it is used in `vhost-user-backend/src/handler.rs`:

```rust
impl<T: VhostUserBackend> Drop for VhostUserHandler<T> {
    fn drop(&mut self) {
        // Signal all working threads to exit.
        self.send_exit_event();

        for thread in self.worker_threads.drain(..) {
            if let Err(e) = thread.join() {
                error!("Error in vring worker: {:?}", e);
            }
        }
    }
}
```

So I really don't understand why every device needs to override it...

BTW for now I think we should do what @uran0sH did, and maybe fix later, because now it really seems that vhost-user-backed used it for both notify and consume.

@germag WDYT?

germag: Agreed, let's keep the `Option<(EventConsumer, EventNotifier)>`, but I think we should add a TODO/FIXME comment so we don't forget we need to refine this in the future

uran0sH:
> > @uran0sH please check the CI, there some clippy warning to fix, but also some tests blocked for more than 60s. Should I restart the CI or it's an issue we need to fix?
> 
> I think it's an issue

Fix this issue by changing this line to into_raw_fd() (https://github.com/rust-vmm/vhost/pull/308/files#diff-e1c9dc7802234bc118de8314ec9cacc98066eea325a6f14ff5887ece3466d5a3R96). Because when we use pipe, using as_raw_fd will cause the consumer's fd to be closed. So we need to take the ownership of this fd.

vhost-user-backend/src/vring.rs
```diff
@@ -15,7 +15,8 @@ use std::sync::{Arc, Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuar
 
 use virtio_queue::{Error as VirtQueError, Queue, QueueT};
 use vm_memory::{GuestAddress, GuestAddressSpace, GuestMemoryAtomic, GuestMemoryMmap};
-use vmm_sys_util::eventfd::EventFd;
+use vmm_sys_util::event::{EventConsumer, EventNotifier};
+// use vmm_sys_util::eventfd::EventFd;
```
stefano-garzarella: To remove

vhost-user-backend/tests/vhost-user-server.rs
```diff
@@ -105,10 +108,11 @@ impl VhostUserBackendMut for MockVhostBackend {
...
     }
 
-    fn exit_event(&self, _thread_index: usize) -> Option<EventFd> {
-        let event_fd = EventFd::new(0).unwrap();
-
-        Some(event_fd)
+    fn exit_event(&self, _thread_index: usize) -> Option<(EventConsumer, EventNotifier)> {
+        Some(
+            new_event_consumer_and_notifier(EventFlag::empty())
+                .expect("Failed to new EventConsumer and EventNotifier"),
```
stefano-garzarella:
```suggestion
                .expect("Failed to create EventConsumer and EventNotifier"),
```

stefano-garzarella:
> This pr([rust-vmm/vmm-sys-util#246](https://github.com/rust-vmm/vmm-sys-util/pull/246)) seems cause the ci failed:
> 
> ```
> thread 'vhost_user::connection::tests::send_fd' panicked at vhost/src/vhost_user/connection.rs:763:50:
> called `Result::unwrap()` on an `Err` value: SocketRetry(Os { code: 105, kind: Uncategorized, message: "No buffer space available" })
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> So I used the version before this PR was merged

@uran0sH please open an issue ASAP on the project, we are preparing a release

germag (approved):
@roypat @stefano-garzarella any thoughts?

PR #311, opened 2025-07-30
vhost: `impl From<UnixListener> for Listener`

alyssais:
### Summary of the PR

It was already possible to do this as follows:

```rust
unsafe { Listener::from_raw_fd(listener.into_raw_fd()) }
```

This just factors the `From` implementation out of the `FromRawFd` implementation so it's possible to do the conversion without an unnecessary `unsafe` in application code.

### Requirements

requirements:

  test. <ins>(Covered by the existing test for `FromRawFd`.)</ins>

### DIFF
--- a/vhost/src/vhost_user/connection.rs
+++ b/vhost/src/vhost_user/connection.rs
@@ -86,10 +86,13 @@ impl AsRawFd for Listener {
 
 impl FromRawFd for Listener {
     unsafe fn from_raw_fd(fd: RawFd) -> Self {
-        Listener {
-            fd: UnixListener::from_raw_fd(fd),
-            path: None,
-        }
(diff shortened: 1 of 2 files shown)


vhost/src/vhost_user/connection.rs
```diff
@@ -86,10 +86,13 @@ impl AsRawFd for Listener {
...
-            fd: UnixListener::from_raw_fd(fd),
-            path: None,
-        }
+        Self::from(UnixListener::from_raw_fd(fd))
+    }
+}
+
+impl From<UnixListener> for Listener {
+    fn from(fd: UnixListener) -> Self {
+        Listener { fd, path: None }
```
epilys: Nit:

```suggestion
        Self { fd, path: None }
```

epilys (approved):

PR #321, opened 2025-09-09
vhost_user: don't take ownership of `Listener`

alyssais:
### Summary of the PR

The vhost-device devices all call `VhostUserDaemon::serve` in a loop, to handle reconnections.  This is not ideal, because a new listener is created each loop iteration, which means that each time, the old socket is unlinked and a new one is created.  This means that there's a potential race where a frontend attempts to connect to the backend before the new socket is created.

A more robust way to achieve this would be to have the devices create their own listeners, and pass the same one to `VhostUserDaemon::start` on each loop iteration, instead of letting `VhostUserDaemon::serve` create it repeatedly.  This was not previously possible though, because `VhostUserDaemon::start` consumed the listener, even though it didn't need to.

Because it's now possible to call `VhostUserDaemon::start` multiple times with the same socket, I've removed the TODO about handling reconnection.  If the TODO was supposed to capture more than that, I can put it back (ideally with clearer phrasing).

### Requirements

requirements:

  test.

### DIFF
--- a/vhost/src/vhost_user/backend.rs
+++ b/vhost/src/vhost_user/backend.rs
@@ -10,16 +10,16 @@ use super::message::*;
 use super::{BackendReqHandler, Result, VhostUserBackendReqHandler};
 
 /// Vhost-user backend side connection listener.
-pub struct BackendListener<S: VhostUserBackendReqHandler> {
-    listener: Listener,
+pub struct BackendListener<'a, S: VhostUserBackendReqHandler> {
+    listener: &'a mut Listener,
     backend: Option<Arc<S>>,
 }
 
 /// Sets up a listener for incoming frontend connections, and handles construction
 /// of a Backend on success.
-impl<S: VhostUserBackendReqHandler> BackendListener<S> {
+impl<'a, S: VhostUserBackendReqHandler> BackendListener<'a, S> {
     /// Create a unix domain socket for incoming frontend connections.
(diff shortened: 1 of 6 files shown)


vhost/src/vhost_user/backend.rs
stefano-garzarella, line 1: Not related to this PR, I mean partially, but should we avoid `self.backend.take().unwrap()` in `BackendListener::accept()`, returning an error instead?

(of course I'm not asking you to fix that here)

stefano-garzarella (commented):
@alyssais LGTM, thanks for that!

> Because it's now possible to call VhostUserDaemon::start multiple times with the same socket, I've removed the TODO about handling reconnection. If the TODO was supposed to capture more than that, I can put it back (ideally with clearer phrasing).

mmm, not sure If we intended to call start multiple times simultaneously, but I don't think so, I think it's okay to remove it.

vhost/src/vhost_user/backend.rs
alyssais: I'm not sure — it looks like `BackendListener` is also designed to handle a single request, and I don't understand why…

epilys (approved):
Makes more sense like this.

vhost/src/vhost_user/backend.rs
stefano-garzarella: yep, right! I mean a second call to `accept()` will panic IIUC, but yeah we can fix in another PR.

stefano-garzarella (approved):

Issue #352, opened 2026-04-07
vhost-user-backend: Allow disabling migration support


DemiMarie:
The vhost-user-backend crate requires that a specific bitmap is used.  Cloud Hypervisor's block code will also require a specific bitmap as part of the fix for a soundness bug (cloud-hypervisor/cloud-hypervisor#7955) which was almost a security vulnerability.  Making the block code generic over the type of bitmap is too complex and likely has negative compile-time impacts.

Therefore, I'm going to be removing this support.  If the backend allowed using any bitmap (with no migration support), this would be better.

germag:
hi, I'm not really follow what you mean, I think I'll understand better with the PR.

> The vhost-user-backend crate requires that a specific bitmap is used

Not really, the first version I wrote was intended for the vm-memory crate, but since the vhost-crate is the only one using it we decided to keep it inside the vhost-user. But, the requirement comes from the vm-memory, any bitmap must implement the Bitmap trait

DemiMarie:
> hi, I'm not really follow what you mean, I think I'll understand better with the PR.

`vhost-user-backend` requries `BitmapMmapRegion`.  This requires locking that makes it slower than the `AtomicBitmap` used elsewhere in Cloud Hypervisor.  Allowing either to be used would require that the whole block layer be made generic over the bitmap.

germag:
> vhost-user-backend requries BitmapMmapRegion. This requires locking that makes it slower than the AtomicBitmap used elsewhere in Cloud Hypervisor. Allowing either to be used would require that the whole block layer be made generic over the bitmap.

Ok, I see, yes we need the lock because the bitmap is only set before migration. But, after setting the bitmap we only get a read lock, so the lock will not be under contention. Did you measure a significant performance degradation?, if so, can we use a different lock implementation where read-only access has a lesser impact.

Still not sure about how you propose to fix it

DemiMarie:
I’d like to remove the requirement for a specific bitmap if live migration is not in use.  The vhost-user backends in Cloud Hypervisor are for testing only.

germag:
> I’d like to remove the requirement for a specific bitmap if live migration is not in use. The vhost-user backends in Cloud Hypervisor are for testing only.

why not just add a feature and use () as bitmap then?, since () already implements all the required traits?

DemiMarie:
[...]

PR #353, opened 2026-04-14
vhost-user-backend: Allow AtomicBitmap

DemiMarie:
### Summary of the PR

https://github.com/cloud-hypervisor/cloud-hypervisor/pull/7971 changes the block layer to require that AtomicBitmap is used.  This breaks the vhost-user server.  Since this server is only for testing, making the whole block layer generic is undesirable.

Allow AtomicBitmap to be used when migration is not needed.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-user-backend/CHANGELOG.md
+++ b/vhost-user-backend/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## [Unreleased]
 
 ### Added
+- Support for using AtomicBitmap (from vm_memory)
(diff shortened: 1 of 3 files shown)


germag:
Hi,

Is this PR related to #352 ?

DemiMarie:
@germag Yes

germag:
> @germag Yes

Ok, so should be close then?

PR #355, opened 2026-04-21
vhost-user-backend: add connection shutdown handle

vishwanath1306:
### Summary of the PR
This commit was needed because the daemon had no reliable way to shut down while a client connection was still alive.  Once the request thread entered the normal vhost-user receive path, it could stay blocked waiting for the next messageand only exit on peer disconnect, which made teardown depend on external timeouts instead of explicit control from the caller. In practice that turned intended shutdown into long hangs during VM teardown and caused spurious test and production timeouts. The change adds an explicit shutdown path that wakes any in-flight socket operation, lets the daemon exit promptly through its existing error path, and clears the per-connection state so later start/wait cycles behave correctly.
### Requirements

requirements:

  test.

### DIFF
--- a/vhost-user-backend/CHANGELOG.md
+++ b/vhost-user-backend/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- [[#355]](https://github.com/rust-vmm/vhost/pull/355) Add an explicit shutdown handle for active daemon connections.
 ### Changed
 ### Deprecated
 ### Fixed
--- a/vhost-user-backend/src/lib.rs
+++ b/vhost-user-backend/src/lib.rs
@@ -9,7 +9,10 @@
 extern crate log;
 
 use std::fmt::{Display, Formatter};
+use std::net::Shutdown;
+use std::os::unix::net::UnixStream;
 use std::path::Path;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::thread;
 
@@ -90,6 +93,27 @@ impl Display for Error {
 /// Result of vhost-user daemon operations.
 pub type Result<T> = std::result::Result<T, Error>;
 
+struct ConnectionState {
+    conn: UnixStream,
+    shutdown_requested: AtomicBool,
+}
+
+/// Thread-safe handle to request shutdown of a running [`VhostUserDaemon`].
+///
+/// Safe to call multiple times or after the connection has already closed.
+#[derive(Clone)]
+pub struct ShutdownHandle {
+    state: Arc<ConnectionState>,
+}
+
+impl ShutdownHandle {
+    /// Request the daemon to shut down.
+    pub fn shutdown(&self) {
+        self.state.shutdown_requested.store(true, Ordering::Release);
+        let _ = self.state.conn.shutdown(Shutdown::Both);
+    }
+}
+
 /// Implement a simple framework to run a vhost-user service daemon.
 ///
 /// This structure is the public API the backend is allowed to interact with in order to run
@@ -98,6 +122,7 @@ pub struct VhostUserDaemon<T: VhostUserBackend> {
     name: String,
     handler: Arc<Mutex<VhostUserHandler<T>>>,
     main_thread: Option<thread::JoinHandle<Result<()>>>,
+    conn_state: Option<Arc<ConnectionState>>,
 }
 
 impl<T> VhostUserDaemon<T>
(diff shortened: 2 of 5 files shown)


sboeuf:
@vishwanath1306 I've pushed a few fixes/enhancements on top of your branch, and also made sure the CI was passing. These changes should be squashed into your commit, I just wanted to highlight them so you could review them.

sboeuf (dismissed):

stefano-garzarella (changes requested):
Please squash all of them before merging...

sboeuf:
> Please squash all of them before merging...

Hi @stefano-garzarella, yes of course these need to be squashed :) , I just wanted @vishwanath1306 to clearly identify the fixes I have added to his original commit.

stefano-garzarella:
@sboeuf yep, but since you approved, I was worried about auto-merge. I really suggest clicking on the `Approve` button when the PR is ready to be merged.

sboeuf:
> @sboeuf yep, but since you approved, I was worried about auto-merge. I really suggest clicking on the `Approve` button when the PR is ready to be merged.

you're right I'm sorry! I've just removed the approval!

vhost-user-backend/src/lib.rs
```diff
@@ -192,14 +260,44 @@ where
...
+        let joined = handle.join();
+        let shutdown_requested = self
+            .conn_state
+            .as_ref()
+            .is_some_and(|s| s.shutdown_requested.load(Ordering::Acquire));
+        self.reset_connection_state();
+
+        match joined.map_err(Error::WaitDaemon)? {
+            Ok(()) => Ok(()),
+            Err(error) if Self::is_expected_wait_error(&error, shutdown_requested) => Ok(()),
```
rbradford: This extra method is unnecessary.

rbradford: I feel like this could be racy - instead we should check the atomic directly in the match statement.

rbradford (commented):
Can you add a double shutdown test since the documentation says it should work.

sboeuf (dismissed):
LGTM!

vhost-user-backend/src/lib.rs
vishwanath1306: Removed it

vishwanath1306: Made this change

vishwanath1306:
> Can you add a double shutdown test since the documentation says it should work.


Added a new test in my commit

vhost-user-backend/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Add an explicit shutdown handle for active daemon connections.
```
stefano-garzarella: Please follow the style of all other entries in this changelog

vhost/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Add socket-cloning helpers for backend request handlers.
```
stefano-garzarella: Ditto

stefano-garzarella (changes requested):

stefano-garzarella:
@vishwanath1306 please resolve all the conversations you fixed. Only PR author and code owners can do that, so it's a good practice that PR author resolve them when they fixed it.

vhost-user-backend/src/lib.rs
```diff
@@ -124,26 +149,45 @@ where
...
+                        break Err(e);
+                    }
+                };
+                let _ = thread_state.conn.shutdown(Shutdown::Both);
+                result
             })
             .map_err(Error::StartDaemon)?;
 
+        self.conn_state = Some(state);
         self.main_thread = Some(handle);
```
stefano-garzarella, line 191: I think this is pre-existing, so not blocking for this PR, but here IIUC we are overwriting those fields, so what happens if the caller calls `start_deamon()` when a deamon is already started? Should we avoid this, e.g. checking `self.main_thread.is_some()` on top, returning an error?

Again, this is pre-existing, so maybe better to fix in another PR, but discovered while reviewing this PR.

stefano-garzarella (approved):
LGTM, I have a comment partially unrelated.

vhost-user-backend/src/lib.rs
vishwanath1306: I will send another PR for this by end of day.

rbradford (approved):

Issue #349, opened 2026-03-24
Use getsockopt to check if socket is AF_UNIX


DemiMarie:
One can use getsockopt to validate that an incoming FD is, in fact, an AF_UNIX socket.

stefano-garzarella:
@DemiMarie can you elaborate a bit more how to use `getsockopt` to do this kind of validation, so newcomers can use this issue as good first issue to fix.

DemiMarie:
@stefano-garzarella Use `getsockopt(SOL_SOCKET, SO_DOMAIN)` to get the domain of the socket, and `getsockopt(SOL_SOCKET, SO_TYPE)` to get its type.  It should also be possible to use `getsockopt(SOL_SOCKET, SO_PROTOCOL)` to get the protocol, but it will always be 0 for `AF_UNIX` sockets.

germag:
> that an incoming FD is, in fact, an AF_UNIX socket.

Can you elaborate this?, which incoming FD?, because that is not always the case for vhost-user messages

DemiMarie:
> > that an incoming FD is, in fact, an AF_UNIX socket.
> 
> Can you elaborate this?, which incoming FD?, because that is not always the case for vhost-user messages

Specifically the reply socket for back-to-front requests.

amniik:
I will work on this and open a PR.

PR #360, opened 2026-06-07
vhost-user: validate AF_UNIX SOCK_STREAM socket file descriptor

amniik:
Validate that the file descriptor received through SET_BACKEND_REQ_FD is an AF_UNIX SOCK_STREAM socket before constructing UnixStream from it.

### Summary of the PR

The vhost-user backend currently assumes that the file descriptor received
through `SET_BACKEND_REQ_FD` is a Unix domain stream socket and constructs a
`UnixStream` from it without validation.

This change validates the received file descriptor before it is used by
checking its socket properties via `getsockopt()`. The validation ensures that
the descriptor corresponds to an `AF_UNIX` `SOCK_STREAM` socket, which is the
expected type for the backend request channel according to the vhost-user
protocol.

The PR also adds tests covering valid and invalid file descriptors.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost/src/vhost_user/backend_req_handler.rs
+++ b/vhost/src/vhost_user/backend_req_handler.rs
@@ -3,6 +3,7 @@
 
 use std::fs::File;
 use std::mem;
+use std::os::fd::{AsFd, BorrowedFd};
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 use std::os::unix::net::UnixStream;
 use std::slice;
@@ -852,9 +853,14 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
 
     fn set_backend_req_fd(&mut self, files: Option<Vec<File>>) -> Result<()> {
         let file = take_single_file(files).ok_or(Error::InvalidMessage)?;
-        // SAFETY: Safe because we have ownership of the files that were
-        // checked when received. We have to trust that they are Unix sockets
-        // since we have no way to check this. If not, it will fail later.
+        // Validate that the received file descriptor is an AF_UNIX SOCK_STREAM
+        // socket as required by the vhost-user backend request channel.
+        validate_unix_stream_socket_fd(file.as_fd())?;
+        // SAFETY:
+        // Ownership of the file descriptor is transferred from `File` via
+        // `into_raw_fd()`, ensuring this code is the sole owner of the FD.
+        // The descriptor has been validated to be an AF_UNIX SOCK_STREAM socket,
+        // so it is safe to construct a `UnixStream` from it.
         let sock = unsafe { UnixStream::from_raw_fd(file.into_raw_fd()) };
         let backend = Backend::from_stream(sock);
         self.backend.set_backend_req_fd(backend);
@@ -863,9 +869,14 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
 
     fn set_gpu_socket(&mut self, files: Option<Vec<File>>) -> Result<()> {
         let file = take_single_file(files).ok_or(Error::InvalidMessage)?;
-        // SAFETY: Safe because we have ownership of the files that were
-        // checked when received. We have to trust that they are Unix sockets
-        // since we have no way to check this. If not, it will fail later.
+        // Validate that the received file descriptor is an AF_UNIX SOCK_STREAM
+        // socket as required by the vhost-user GPU channel.
+        validate_unix_stream_socket_fd(file.as_fd())?;
+        // SAFETY:
+        // Ownership of the file descriptor is transferred from `File` via
+        // `into_raw_fd()`, ensuring this code is the sole owner of the FD.
+        // The descriptor has been validated to be an AF_UNIX SOCK_STREAM socket,
+        // so it is safe to construct a `UnixStream` from it.
         let sock = unsafe { UnixStream::from_raw_fd(file.into_raw_fd()) };
         let gpu_backend = GpuBackend::from_stream(sock);
         self.backend.set_gpu_socket(gpu_backend)
@@ -1036,8 +1047,52 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
     }
 }
 
+// Retrieve a SOL_SOCKET socket option value using `getsockopt`.
+fn get_socket_opt(fd: BorrowedFd<'_>, opt: libc::c_int) -> std::io::Result<libc::c_int> {
+    let mut value: libc::c_int = 0;
+    let mut len = std::mem::size_of::<libc::c_int>() as libc::socklen_t;
+
+    // SAFETY:
+    // - `value` and `len` point to valid writable memory.
+    // - `len` is initialized to `size_of::<libc::c_int>()`, matching the size
+    //   of `value`, so `getsockopt()` will not write past the end of `value`.
+    // - `fd.as_raw_fd()` is valid because `BorrowedFd` guarantees a live file
+    //   descriptor for the duration of the call.
+    let rc = unsafe {
+        libc::getsockopt(
+            fd.as_raw_fd(),
+            libc::SOL_SOCKET,
+            opt,
+            std::ptr::addr_of_mut!(value).cast::<libc::c_void>(),
+            std::ptr::addr_of_mut!(len),
+        )
+    };
+
+    if rc == -1 {
+        Err(std::io::Error::last_os_error())
+    } else {
+        Ok(value)
+    }
+}
+
+// Validate that the given file descriptor is an AF_UNIX SOCK_STREAM socket.
+fn validate_unix_stream_socket_fd(fd: BorrowedFd<'_>) -> Result<()> {
+    let domain = get_socket_opt(fd, libc::SO_DOMAIN).map_err(Error::InvalidSocketFd)?;
+    if domain != libc::AF_UNIX {
+        return Err(Error::NotUnixSocket);
+    }
+
+    let sock_type = get_socket_opt(fd, libc::SO_TYPE).map_err(Error::InvalidSocketFd)?;
+    if sock_type != libc::SOCK_STREAM {
+        return Err(Error::NotStreamSocket);
+    }
+
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
+    use std::net::TcpListener;
     use std::os::unix::io::AsRawFd;
 
     use super::*;
@@ -1158,5 +1213,24 @@ mod tests {
             let _ = frontend_endpoint.send_message(&hdr, &VhostUserEmpty, None);
         });
         assert!(handler.handle_request().is_err());
+    }
+
+    #[test]
+    fn test_validate_unix_stream_socket_fd() {
+        // Valid AF_UNIX SOCK_STREAM socket
+        let (sock1, _sock2) = UnixStream::pair().unwrap();
+        assert!(validate_unix_stream_socket_fd(sock1.as_fd()).is_ok());
+
+        // AF_INET socket (wrong domain)
+        let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+        assert!(validate_unix_stream_socket_fd(listener.as_fd()).is_err());
+
+        // Non-socket file descriptor
+        let file = File::open("/dev/null").unwrap();
+        assert!(validate_unix_stream_socket_fd(file.as_fd()).is_err());
+
+        // AF_UNIX but SOCK_DGRAM (wrong type)
+        let dgram = std::os::unix::net::UnixDatagram::unbound().unwrap();
+        assert!(validate_unix_stream_socket_fd(dgram.as_fd()).is_err());
     }
 }
--- a/vhost/src/vhost_user/mod.rs
+++ b/vhost/src/vhost_user/mod.rs
@@ -68,6 +68,12 @@ pub enum Error {
     InactiveOperation(VhostUserProtocolFeatures),
     /// Invalid message format, flag or content.
     InvalidMessage,
+    /// Invalid `AF_UNIX SOCK_STREAM` socket file descriptor.
+    InvalidSocketFd(std::io::Error),
+    /// Socket is not `AF_UNIX`
+    NotUnixSocket,
+    /// Socket is not `SOCK_STREAM`
+    NotStreamSocket,
     /// Only part of a message have been sent or received successfully
     PartialMessage,
     /// The peer disconnected from the socket.
@@ -110,6 +116,9 @@ impl std::fmt::Display for Error {
                 write!(f, "inactive protocol operation: {}", bits.bits())
             }
             Error::InvalidMessage => write!(f, "invalid message"),
+            Error::InvalidSocketFd(e) => write!(f, "invalid AF_UNIX SOCK_STREAM socket fd: {e}"),
+            Error::NotUnixSocket => write!(f, "socket is not AF_UNIX"),
+            Error::NotStreamSocket => write!(f, "socket is not SOCK_STREAM"),
             Error::PartialMessage => write!(f, "partial message"),
             Error::Disconnected => write!(f, "peer disconnected"),
             Error::OversizedMsg => write!(f, "oversized message"),
@@ -157,6 +166,9 @@ impl Error {
             Error::InvalidParam | Error::InvalidOperation(_) => false,
             Error::InactiveFeature(_) | Error::InactiveOperation(_) => false,
             Error::InvalidMessage | Error::IncorrectFds | Error::OversizedMsg => false,
+            Error::InvalidSocketFd(_) => false,
+            Error::NotUnixSocket => false,
+            Error::NotStreamSocket => false,
             Error::SocketError(_) | Error::SocketConnect(_) => false,
             Error::FeatureMismatch => false,
             Error::ReqHandlerError(_) => false,
@@ -268,6 +280,7 @@ mod tests {
     use message::VhostUserSharedMsg;
     use std::fs::File;
     use std::os::unix::io::AsRawFd;
+    use std::os::unix::net::UnixStream;
     use std::path::{Path, PathBuf};
     use std::sync::{Arc, Barrier, Mutex};
     use std::thread;
@@ -490,6 +503,7 @@ mod tests {
         assert_eq!(num, 2);
 
         let eventfd = vmm_sys_util::eventfd::EventFd::new(0).unwrap();
+        let (tx, _) = UnixStream::pair().unwrap();
         let mem = [VhostUserMemoryRegionInfo::new(
             0,
             0x10_0000,
@@ -510,7 +524,7 @@ mod tests {
         assert_eq!(offset, 0x100);
         assert_eq!(&reply_payload, &[0xa5; 4]);
 
-        frontend.set_backend_request_fd(&eventfd).unwrap();
+        frontend.set_backend_request_fd(&tx).unwrap();
         frontend.set_vring_enable(0, true).unwrap();
 
         frontend
--- a/vhost/CHANGELOG.md
+++ b/vhost/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 - [[#355]](https://github.com/rust-vmm/vhost/pull/355) Add socket-cloning helpers for backend request handlers.
+- [[#360]](https://github.com/rust-vmm/vhost/pull/360) Validate AF_UNIX SOCK_STREAM socket file descriptors and 
+  add Error::InvalidSocketFd, Error::NotUnixSocket and Error::NotStreamSocket.
 ### Changed
 ### Deprecated
 ### Fixed


amniik:
This PR addresses issue: https://github.com/rust-vmm/vhost/issues/349

epilys:
Put:

```text
Fixes #349 ("Use getsockopt to check if socket is AF_UNIX")
```

In commit message and github will close the issue automatically when it's merged.

amniik:
> Put:
> 
> ```
> Fixes #349 ("Use getsockopt to check if socket is AF_UNIX")
> ```
> 
> In commit message and github will close the issue automatically when it's merged.

Thanks! I've updated the commit message

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,9 +1040,73 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
...
+    // - `getsockopt` is a well-defined libc syscall that only writes to provided buffers
+    let rc = unsafe {
+        libc::getsockopt(
+            fd,
+            libc::SOL_SOCKET,
+            opt,
+            &mut value as *mut _ as *mut libc::c_void,
+            &mut len,
+        )
+    };
```
arctic-alpaca: `fd` isn't guaranteed to be valid, which means the safety documentation cannot rely on it being valid. Instead you can argue that `getsockopt` handles invalid file descriptors.

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,9 +1040,73 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
     }
 }
 
+// Retrieve a socket option value using getsockopt().
```
arctic-alpaca: Is this (and on functions below) intended not to be a doc comment (`///`)?

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -852,9 +852,13 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
...
-        // checked when received. We have to trust that they are Unix sockets
-        // since we have no way to check this. If not, it will fail later.
+        // Validate that the file descriptor received through
+        // VHOST_USER_SET_BACKEND_REQ_FD is an AF_UNIX SOCK_STREAM socket
+        // before constructing a UnixStream from it.
+        validate_backend_req_fd(file.as_raw_fd())?;
+        // SAFETY: We have validated that the file descriptor is an
+        // AF_UNIX SOCK_STREAM socket. Therefore it is safe to construct
+        // a UnixStream from it and assume correct socket semantics.
         let sock = unsafe { UnixStream::from_raw_fd(file.into_raw_fd()) };
```
arctic-alpaca: The safety reasoning isn't complete since `UnixStream::from_raw_fd` also requires the passed fd to be exclusively owned (https://doc.rust-lang.org/std/io/index.html#io-safety).

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,9 +1040,73 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
...
+
+// Get the socket type (SO_TYPE) of a socket.
+fn socket_type(fd: RawFd) -> std::io::Result<libc::c_int> {
+    get_socket_opt(fd, libc::SO_TYPE)
+}
+
+// Get the protocol (SO_PROTOCOL) associated with a socket.
+fn socket_protocol(fd: RawFd) -> std::io::Result<libc::c_int> {
+    get_socket_opt(fd, libc::SO_PROTOCOL)
+}
```
arctic-alpaca: If this isn't intended to be used elsewhere too, I think it's cleaner to inline it into `validate_backend_req_fd`.

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,9 +1040,73 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
...
+    // SAFETY: This call is safe because:
+    // - `fd` is expected to be a valid file descriptor
+    // - `value` and `len` are valid mutable buffers of correct size
+    // - `getsockopt` is a well-defined libc syscall that only writes to provided buffers
+    let rc = unsafe {
+        libc::getsockopt(
+            fd,
+            libc::SOL_SOCKET,
+            opt,
+            &mut value as *mut _ as *mut libc::c_void,
```
arctic-alpaca: Using `.cast()` instead of `as *mut ...` prevents accidental mutability changes in pointer casts.

Optional, but you can also avoid the `as *mut _` if you go for `(&raw mut value).cast()`.

vhost/src/vhost_user/backend_req_handler.rs
epilys: Private functions do not need rustdoc comments.

vhost/src/vhost_user/backend_req_handler.rs
epilys: the `fd` comes from an owned `std::fs::File`, so it is owned.

vhost/src/vhost_user/backend_req_handler.rs
epilys: Just remove the `fd is expected to be a valid file descriptor` line from the safety comment. It is just an integer so has no effect on the safety of the FFI call.

vhost/src/vhost_user/backend_req_handler.rs
arctic-alpaca: While they may not need rustdoc comments, the functions here already have helpful documentation which currently isn't visible for example on hover in IDEs.

vhost/src/vhost_user/backend_req_handler.rs
arctic-alpaca: Yes and that was part of the safety reasoning before the change, so I think it's reasonable to keep.

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,9 +1040,73 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
     }
 }
 
+// Retrieve a socket option value using getsockopt().
+fn get_socket_opt(fd: RawFd, opt: libc::c_int) -> std::io::Result<libc::c_int> {
```
germag: Let's use this opportunity to start making this more type safe:
`fn get_socket_opt(fd: BorrowedFd...)`  then we can drop all callers safety comments-

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -852,9 +852,13 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
 
     fn set_backend_req_fd(&mut self, files: Option<Vec<File>>) -> Result<()> {
         let file = take_single_file(files).ok_or(Error::InvalidMessage)?;
-        // SAFETY: Safe because we have ownership of the files that were
-        // checked when received. We have to trust that they are Unix sockets
-        // since we have no way to check this. If not, it will fail later.
+        // Validate that the file descriptor received through
+        // VHOST_USER_SET_BACKEND_REQ_FD is an AF_UNIX SOCK_STREAM socket
+        // before constructing a UnixStream from it.
+        validate_backend_req_fd(file.as_raw_fd())?;
```
germag: Using ` validate_backend_req_fd(file.as_fd())?;` we can drop the safety comment

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,9 +1040,73 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
...
+}
+
+// Get the protocol (SO_PROTOCOL) associated with a socket.
+fn socket_protocol(fd: RawFd) -> std::io::Result<libc::c_int> {
+    get_socket_opt(fd, libc::SO_PROTOCOL)
+}
+
+// Validate that the backend request file descriptor is an
+// AF_UNIX SOCK_STREAM socket.
+fn validate_backend_req_fd(fd: RawFd) -> Result<()> {
```
germag: `fn validate_backend_req_fd(fd: BorrowedFd) -> Result<()>`

vhost/src/vhost_user/backend_req_handler.rs
germag: I don't see the purpose of these functions, why not just call `get_sock_op` directly inside `validate_backend_req_fd()`?

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,9 +1040,73 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
...
+
+    let sock_type = socket_type(fd).map_err(|_| Error::InvalidBackendReqFD)?;
+    if sock_type != libc::SOCK_STREAM {
+        return Err(Error::InvalidBackendReqFD);
+    }
+
+    let protocol = socket_protocol(fd).map_err(|_| Error::InvalidBackendReqFD)?;
+    if protocol != 0 {
+        return Err(Error::InvalidBackendReqFD);
+    }
```
germag: I don't see the purpose of these functions, why not just call get_sock_op directly inside validate_backend_req_fd()?

vhost/src/vhost_user/mod.rs
```diff
@@ -110,6 +112,7 @@ impl std::fmt::Display for Error {
                 write!(f, "inactive protocol operation: {}", bits.bits())
             }
             Error::InvalidMessage => write!(f, "invalid message"),
+            Error::InvalidBackendReqFD => write!(f, "invalid backend request fd"),
```
germag: `InvalidBackendReqFd`  for consistency with the rest of errors

vhost/src/vhost_user/backend_req_handler.rs
germag: This SAFETY comment is outdated now, check the rest.

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,9 +1040,73 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
...
+}
+
+// Get the socket type (SO_TYPE) of a socket.
+fn socket_type(fd: RawFd) -> std::io::Result<libc::c_int> {
+    get_socket_opt(fd, libc::SO_TYPE)
+}
+
+// Get the protocol (SO_PROTOCOL) associated with a socket.
+fn socket_protocol(fd: RawFd) -> std::io::Result<libc::c_int> {
+    get_socket_opt(fd, libc::SO_PROTOCOL)
```
germag: Do we really need this?, it always 0, I think checking for domain and type should be enough

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1159,4 +1227,19 @@ mod tests {
...
+
+        let eventfd = EventFd::new(0).unwrap();
+        assert!(validate_backend_req_fd(eventfd.as_raw_fd()).is_err());
+
+        let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+        assert!(validate_backend_req_fd(listener.as_raw_fd()).is_err());
+
+        let file = File::open("/dev/null").unwrap();
+        assert!(validate_backend_req_fd(file.as_raw_fd()).is_err());
+    }
```
germag:
```
  let dgram = std::os::unix::net::UnixDatagram::unbound().unwrap();
  assert!(validate_backend_req_fd(dgram.as_raw_fd()).is_err());
``` 
maybe?

vhost/src/vhost_user/backend_req_handler.rs
germag: Future note (not for this PR): We should have a `BackendReqFd` that must be check on construction

vhost/src/vhost_user/backend_req_handler.rs
amniik: Thanks, I've rewritten it with BorrowedFd.

vhost/src/vhost_user/backend_req_handler.rs
amniik: I kept regular comments here to stay consistent with the existing style, since other private helper functions are not documented with rustdoc comments. This can be revisited if the project's style changes in the future.

vhost/src/vhost_user/backend_req_handler.rs
amniik: OK, I have added this info.

vhost/src/vhost_user/backend_req_handler.rs
amniik: Yes, I'v rewritten it.

vhost/src/vhost_user/backend_req_handler.rs
amniik: Good idea, I’ve applied it.

vhost/src/vhost_user/backend_req_handler.rs
amniik: Thanks, I'v applied it.

vhost/src/vhost_user/backend_req_handler.rs
amniik: I'v applied it.

vhost/src/vhost_user/backend_req_handler.rs
amniik: I'v applied it.

vhost/src/vhost_user/backend_req_handler.rs
amniik: I agree. The helper functions are currently only used here, so I’ll inline the get_socket_opt calls in validate_backend_req_fd for now to keep the code simpler. We can reintroduce them later if they are needed elsewhere.

vhost/src/vhost_user/backend_req_handler.rs
amniik: I agree. The helper functions are currently only used here, so I’ll inline the get_socket_opt calls in validate_backend_req_fd for now to keep the code simpler. We can reintroduce them later if they are needed elsewhere.

vhost/src/vhost_user/mod.rs
amniik: Thanks. I'v fixed it.

vhost/src/vhost_user/backend_req_handler.rs
amniik: I agree. This check was suggested in the issue as optional, so I’ve removed it.

vhost/src/vhost_user/backend_req_handler.rs
amniik: Yes, I'v added it.

vhost/src/vhost_user/backend_req_handler.rs
amniik: Do you think it would make sense to generalize validate_backend_req_fd and InvalidBackendReqFD to something like validate_unix_stream_socket_fd and InvalidSocketFD, since the validation is now used by both the backend request FD and the GPU socket path?

germag: Yes, I agree that now the name is incorrect, let's rename it, your proposed names are fine.

germag: You also should update the commit, PR and changelog messages

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,8 +1047,51 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
...
+
+// Validate that the backend request file descriptor is an
+// AF_UNIX SOCK_STREAM socket.
+fn validate_backend_req_fd(fd: BorrowedFd<'_>) -> Result<()> {
+    let domain = get_socket_opt(fd, libc::SO_DOMAIN).map_err(|_| Error::InvalidBackendReqFd)?;
+    if domain != libc::AF_UNIX {
+        return Err(Error::InvalidBackendReqFd);
+    }
+
+    let sock_type = get_socket_opt(fd, libc::SO_TYPE).map_err(|_| Error::InvalidBackendReqFd)?;
```
germag: Maybe could be useful (specially for debugging):
```
.map_err(|e| Error::InvalidBackendReqFd(e))?;
```

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,8 +1047,51 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
...
+    // - `value` and `len` are valid writable buffers
+    // - `fd.as_raw_fd()` is valid because BorrowedFd guarantees a live FD
+    // - libc::getsockopt only writes to provided buffers
+    let rc = unsafe {
+        libc::getsockopt(
+            fd.as_raw_fd(),
+            libc::SOL_SOCKET,
+            opt,
+            (&raw mut value).cast::<libc::c_void>(),
+            (&raw mut len).cast::<libc::socklen_t>(),
```
germag: Not an issue, just a comment

the `&raw ...` is available since rust 1.82, should we add `rust-version = "1.82"` to `Cargo.toml`?, although Idk if there is something else that requires an even newer version. But at least we need to take this into account

germag: This `(&raw mut len).cast::<libc::socklen_t>(),` can be just `&raw mut len`  (or `std::ptr::addr_of_mut!(len)`) there is already a cast before `let mut len = std::mem::size_of::<libc::c_int>() as libc::socklen_t;`

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,8 +1047,51 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
...
 
+// Retrieve a socket option value using `getsockopt`.
+fn get_socket_opt(fd: BorrowedFd<'_>, opt: libc::c_int) -> std::io::Result<libc::c_int> {
+    let mut value: libc::c_int = 0;
+    let mut len = std::mem::size_of::<libc::c_int>() as libc::socklen_t;
+
+    // SAFETY:
+    // - `value` and `len` are valid writable buffers
+    // - `fd.as_raw_fd()` is valid because BorrowedFd guarantees a live FD
+    // - libc::getsockopt only writes to provided buffers
```
germag: It should be noted that `getsockopt`, in these cases, writes `std::mem::size_of::<libc::c_int>()` bytes, and that is why `len` is initialized on that value.

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -1036,8 +1047,51 @@ impl<S: VhostUserBackendReqHandler> AsRawFd for BackendReqHandler<S> {
     }
 }
 
+// Retrieve a socket option value using `getsockopt`.
+fn get_socket_opt(fd: BorrowedFd<'_>, opt: libc::c_int) -> std::io::Result<libc::c_int> {
```
germag: I guess, it should be something like "Retrieve a SOL_SOCKET socket option value  using `getsockopt(fd, SOL_SOCKET...)`.", since `libc::SOL_SOCKET` is hardcoded

vhost/src/vhost_user/backend_req_handler.rs
amniik: Yes I'v applied it.

vhost/src/vhost_user/backend_req_handler.rs
amniik: Yes, I agree. I also added Error::NotUnixSocket and Error::NotStreamSocket for more precise error reporting.

vhost/src/vhost_user/backend_req_handler.rs
amniik: Yes I'v updated the comment.

vhost/src/vhost_user/backend_req_handler.rs
amniik: Yes I'v added this info.

vhost/src/vhost_user/backend_req_handler.rs
amniik: I didn’t see any rust-version specified in the project, so for compatibility I’ve switched back to the older form.

amniik: Yes, I'v applied this.

vhost/src/vhost_user/backend_req_handler.rs
amniik: For compatibility reason, we kept old form.

epilys: You can use `.cast()` even without `&raw mut`, so there's no compatibility problem here.

amniik: Yes, I'v changed it.

vhost/src/vhost_user/mod.rs
```diff
@@ -68,6 +68,12 @@ pub enum Error {
     InactiveOperation(VhostUserProtocolFeatures),
     /// Invalid message format, flag or content.
     InvalidMessage,
+    /// Invalid AF_UNIX SOCK_STREAM socket file descriptor.
+    InvalidSocketFd(std::io::Error),
+    /// Socket is not AF_UNIX
+    NotUnixSocket,
+    /// Socket is not SOCK_STREAM
+    NotStreamSocket,
```
epilys:
```suggestion
    /// Invalid `AF_UNIX SOCK_STREAM` socket file descriptor.
    InvalidSocketFd(std::io::Error),
    /// Socket is not `AF_UNIX`
    NotUnixSocket,
    /// Socket is not `SOCK_STREAM`
    NotStreamSocket,
```

amniik: I'v applied it.

epilys (approved):

germag (approved):

vhost/src/vhost_user/backend_req_handler.rs
arctic-alpaca: Sorry for not responding, I was on vacation. Thanks for taking my comments into account!

PR #316, opened 2025-08-19
Use mio to replace Epoll

uran0sH:
### Summary of the PR

Epoll is linux-specific. So we use mio, which is a cross-platform event notification, to replace Epoll.

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [ x ] All commits in this PR have Signed-Off-By trailers (with
  `git commit -s`), and the commit message has max 60 characters for the
  summary and max 75 characters for each description line.
- [ x ] All added/changed functionality has a corresponding unit/integration
  test.
- [ x ] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [ x ] Any newly added `unsafe` code is properly documented.

PR #362, opened 2026-07-03
epoll to mio migration

eric-14:
### Summary of the PR

Epoll is linux specific. To support unix systems this PR changes I/O support  to mio. 

PR is also a continuation of https://github.com/rust-vmm/vhost/pull/316.

The PR completes the work needed to move from epoll to mio.

### Requirements

requirements:

[ x ] All commits in this PR have Signed-Off-By trailers (with
git commit -s), and the commit message has max 60 characters for the
[ x ] All added/changed functionality has a corresponding unit/integration
test.
[ x ] All added/changed public-facing functionality has entries in the "Upcoming
[ x ] Any newly added unsafe code is properly documented.

### DIFF
--- a/vhost-user-backend/src/backend.rs
+++ b/vhost-user-backend/src/backend.rs
@@ -29,13 +29,14 @@ use vhost::vhost_user::message::{
 };
 use vhost::vhost_user::Backend;
 use vm_memory::bitmap::Bitmap;
-use vmm_sys_util::epoll::EventSet;
 use vmm_sys_util::event::{EventConsumer, EventNotifier};
 
 use vhost::vhost_user::GpuBackend;
 
 use super::vring::VringT;
 use super::GM;
+
+use crate::EventSet;
 
 /// Trait with interior mutability for vhost user backend servers to implement concrete services.
 ///
@@ -828,7 +829,7 @@ pub mod tests {
 
         let vring = VringRwLock::new(mem, 0x1000).unwrap();
         backend
-            .handle_event(0x1, EventSet::IN, &[vring], 0)
+            .handle_event(0x1, EventSet::Readable, &[vring], 0)
             .unwrap();
 
         backend.reset_device();
--- a/vhost-user-backend/src/backend.rs
+++ b/vhost-user-backend/src/backend.rs
@@ -144,7 +144,7 @@ pub trait VhostUserBackend: Send + Sync {
     /// do with events happening on custom listeners.
     fn handle_event(
         &self,
-        device_event: u16,
+        device_event: usize,
         evset: EventSet,
         vrings: &[Self::Vring],
         thread_id: usize,
@@ -295,7 +295,7 @@ pub trait VhostUserBackendMut: Send + Sync {
     /// do with events happening on custom listeners.
     fn handle_event(
         &mut self,
-        device_event: u16,
+        device_event: usize,
         evset: EventSet,
         vrings: &[Self::Vring],
         thread_id: usize,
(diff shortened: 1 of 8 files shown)


vhost-user-backend/src/backend.rs
```diff
@@ -144,7 +145,7 @@ pub trait VhostUserBackend: Send + Sync {
     /// do with events happening on custom listeners.
     fn handle_event(
         &self,
-        device_event: u16,
+        device_event: usize,
```
jemoreira, line 148: This change will break every existing implementation of the trait. Would it be possible to keep `u16` and cast as needed?

vhost-user-backend/src/backend.rs
```diff
@@ -828,7 +829,7 @@ pub mod tests {
 
         let vring = VringRwLock::new(mem, 0x1000).unwrap();
         backend
-            .handle_event(0x1, EventSet::IN, &[vring], 0)
+            .handle_event(0x1, EventSet::Readable, &[vring], 0)
```
jemoreira, line 832: same idea here: most implementations already expect EventSet::IN (although it's ignored in most cases).

vhost-user-backend/src/event_loop.rs
```diff
@@ -3,70 +3,115 @@
...
-/// Epoll event handler to manage and process epoll events for registered file descriptor.
+#[derive(Debug, Clone, Copy)]
+pub enum EventSet {
+    Readable,
+    Writable,
+    All,
+}
+
+impl EventSet {
+    fn to_interest(self) -> Interest {
```
jemoreira, line 69: nit: Consider implementing `std::convert::Into<Interest>` instead.

vhost-user-backend/src/event_loop.rs
```diff
@@ -3,70 +3,115 @@
...
+        match self {
+            EventSet::Readable => Interest::READABLE,
+            EventSet::Writable => Interest::WRITABLE,
+            EventSet::All => Interest::READABLE | Interest::WRITABLE,
+        }
+    }
+}
+
+fn event_to_event_set(evt: &Event) -> Result<EventSet> {
```
jemoreira, line 78: nit: consider `std::convert::TryFrom<Event>`

vhost-user-backend/src/event_loop.rs
```diff
@@ -112,100 +166,107 @@ where
...
     }
 
-    pub(crate) fn register_event(&self, fd: RawFd, ev_type: EventSet, data: u64) -> Result<()> {
-        self.epoll
-            .ctl(ControlOperation::Add, fd, EpollEvent::new(ev_type, data))
+    pub(crate) fn register_event(
+        &self,
+        fd: RawFd,
+        ev_type: EventSet,
+        queue_idx: usize,
```
jemoreira, line 199: nit: This parameter may not be a queue index, `event_id` might be a better name.

vhost-user-backend/src/handler.rs
```diff
@@ -137,14 +138,17 @@ where
                 }
             }
 
+            let poller = Poll::new()
+                .map_err(VringPollError::PollerCreate)
+                .map_err(VhostUserHandlerError::CreatePollHandler)?;
             let handler = Arc::new(
-                VringEpollHandler::new(backend.clone(), thread_vrings, thread_id)
-                    .map_err(VhostUserHandlerError::CreateEpollHandler)?,
+                VringPollHandler::new(backend.clone(), thread_vrings, thread_id, &poller)
```
jemoreira, line 145: Why is `poller` borrowed here and in the call to `run` instead of simply owned by `VringPollHandler`? Given that nothing else seems to use it.

eric-14: From pr-316 review, ``poller`` was part of ``VringPollHandler`` and required locking for the sync trait. Mio is inherently single threaded. The above implementation presents an alternative where poller is borrowed without object access being synced with a mutex across multiple threads.  
The initial suggestion was made in [PR-316](https://github.com/rust-vmm/vhost/pull/316#issuecomment-4766286925).

PR #368, opened 2026-07-13
Fix docs.rs builds

epilys:
docs.rs builds are failing: <https://docs.rs/crate/vhost-user-backend/0.23.0>

Because PR https://github.com/rust-vmm/vhost/pull/326 did not use the proper `rustc-flags` value for docs.rs builds.

Tested with <https://github.com/syphar/cargo-docs-rs> :

```shell
cargo +nightly docs-rs -p vhost
cargo +nightly docs-rs -p vhost-user-backend
```

### Requirements

requirements:

  test.

### DIFF
--- a/vhost/build.rs
+++ b/vhost/build.rs
@@ -1,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+fn main() {
+    println!("cargo::rerun-if-env-changed=DOCS_RS");
+    if std::env::var("DOCS_RS").is_ok() {
+        println!("cargo::rustc-cfg=RUSTDOC_disable_feature_compat_errors");
+    }
+}
--- a/vhost-user-backend/Cargo.toml
+++ b/vhost-user-backend/Cargo.toml
@@ -10,7 +10,7 @@ license = "Apache-2.0"
 
 [package.metadata.docs.rs]
 all-features = true
-rustc-args  = ['--cfg RUSTDOC_disable_feature_compat_errors']
+rustc-args = ["--cfg", "RUSTDOC_disable_feature_compat_errors"]
 
 [features]
 xen = ["vm-memory/xen", "vhost/xen"]
--- a/vhost-user-backend/build.rs
+++ b/vhost-user-backend/build.rs
@@ -1,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+fn main() {
+    println!("cargo::rerun-if-env-changed=DOCS_RS");
+    if std::env::var("DOCS_RS").is_ok() {
+        println!("cargo::rustc-cfg=RUSTDOC_disable_feature_compat_errors");
+    }
+}
--- a/vhost/Cargo.toml
+++ b/vhost/Cargo.toml
@@ -12,7 +12,7 @@ edition = "2021"
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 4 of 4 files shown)


vhost/build.rs
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+fn main() {
+    if std::env::var("DOCS_RS").is_ok() {
```
stefano-garzarella, line 5: What about adding `println!("cargo::rerun-if-env-changed=DOCS_RS");` ?

vhost/build.rs
stefano-garzarella, line 1: What about adding the same `build.rs` also in `vhost-user-backend` so we can remove the `rustc-args` in Cargo.toml files? (maybe it will also improve the situation with docs build of consumers, e.g. vhost-devices)

Cargo.lock
stefano-garzarella, line 1: nit: I would add two lines to the commit explaining why we're starting to track this, so that new contributors are aware of it.

vhost/build.rs
epilys: Good catch, fixed

vhost/build.rs
epilys: Done, not sure if `rustc-args` should be removed though. It builds without them but for some reason they recommend both? Don't really know what's the best thing to do here.

Cargo.lock
epilys: Done, thanks!

vhost/build.rs
stefano-garzarella: I'm fine also leaving both!

stefano-garzarella (dismissed):

stefano-garzarella:
I just saw that #365 updated `nix` but we didn't merge it yet. BTW I'm fine to keep the update here, I was just checking why dependabot didn't do it.

.gitignore
```diff
@@ -3,4 +3,3 @@
 /target
 .idea
 **/*.rs.bk
-Cargo.lock
```
andreeaflorescu, line 6: The problem with this is that when these crates are used we don't have control over the Cargo.lock. While the build might pass for our CI, it could fail when used by other crates as the Cargo.lock will be ignored.

What was the reason for the failure you encounter? I am trying to understand whether there is another way to fix it.

> It's also recommended by the Rust project:

My read of that FAQ is different. It is saying there are some reasons why you might want to have it versioned, but it also mentions the pitfalls. In the end it's a per project decision. We've been committing Cargo.lock only for binary projects so far.

epilys: > The problem with this is that when these crates are used we don't have control over the Cargo.lock

Yeah, so it shouldn't make a difference to them. This only affects local development.

> What was the reason for the failure you encounter? I am trying to understand whether there is another way to fix it.

Just a local, *very* outdated Cargo.lock.

andreeaflorescu: > Yeah, so it shouldn't make a difference to them. This only affects local development.

It does affect the CI though. That's what concerns me. If we don't have a Cargo.lock file then the CI will run with the dependencies up to date as per `Cargo.toml`.

andreeaflorescu: Are you running in a way that is locking the Cargo.lock file? Otherwise for library crates the Cargo.lock file is overwritten whenever you do a build.

epilys: No, and I don't have the Cargo.lock anymore since it was overwritten. I will remove the commit to unblock the PR since you have an objection

andreeaflorescu (approved):

stefano-garzarella (approved):

PR #369, opened 2026-07-13
vhost-user: add GpuFrontend API for GPU socket communication

dorindabassey:
### Summary of the PR

Add frontend-side API for the vhost-user-gpu protocol socket:
- VhostUserFrontend::set_gpu_socket() sends GPU_SET_SOCKET message
- GpuFrontend reads requests and sends typed responses
- Mirrors existing GpuBackend pattern

Eliminates need for frontends to manually construct GPU protocol headers.

update: 
Added `recv_payload_exact()` to handle partial read - the original `recv_into_buf()` would truncate large payloads (like gfxstream framebuffers) because `recv()` could return fewer bytes than requested.
Also added `send_header()` for some messages like `DMABUF_UPDATE` that need header-only ACK responses.

### Requirements

requirements:

  test.


vhost/src/vhost_user/gpu_frontend_req.rs
```diff
@@ -0,0 +1,227 @@
...
+        node.check_state()?;
+
+        let (hdr, files) = node.sock.recv_header().map_err(io::Error::other)?;
+        let request = hdr.get_code().map_err(io::Error::other)?;
+        let size = hdr.get_size() as usize;
+
+        let (payload, payload_files) = if size > 0 {
+            let (_bytes, payload, files) =
+                node.sock.recv_into_buf(size).map_err(io::Error::other)?;
+            (payload, files)
```
germag: Is it possible to have FD received with the header and the payload? If that is the case, the first `files` gets lost. 

If it is not possible (which I think is the case), I guess a comment will be nice. And maybe also reverse the order of
```rust
let all_files = payload_files.or(files);
```

vhost/src/vhost_user/frontend.rs
```diff
@@ -506,6 +509,13 @@ impl VhostUserFrontend for Frontend {
         node.wait_for_ack(&hdr).map_err(|e| e.into())
     }
 
+    fn set_gpu_socket(&mut self, fd: &dyn AsRawFd) -> Result<()> {
+        let mut node = self.node();
```
germag, line 513: Doesn't this require any protocol feature?

vhost/src/vhost_user/gpu_frontend_req.rs
```diff
@@ -0,0 +1,227 @@
...
+        let mut node = self.node();
+        node.check_state()?;
+
+        let (hdr, files) = node.sock.recv_header().map_err(io::Error::other)?;
+        let request = hdr.get_code().map_err(io::Error::other)?;
+        let size = hdr.get_size() as usize;
+
+        let (payload, payload_files) = if size > 0 {
+            let (_bytes, payload, files) =
+                node.sock.recv_into_buf(size).map_err(io::Error::other)?;
```
germag, line 89: Hmmm, here the code blindly trusts the hdr size and  inside `recv_into_buf(size)` we are doing an allocation: `let mut buf = vec![0u8; buf_size];`, could this be a problem?

Maybe the check should be inside `recv_into_buf()` ?

vhost/src/vhost_user/gpu_frontend_req.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub fn read_message(&self) -> io::Result<(GpuBackendReq, Vec<u8>, Option<Vec<File>>)> {
+        let mut node = self.node();
+        node.check_state()?;
+
+        let (hdr, files) = node.sock.recv_header().map_err(io::Error::other)?;
+        let request = hdr.get_code().map_err(io::Error::other)?;
+        let size = hdr.get_size() as usize;
+
+        let (payload, payload_files) = if size > 0 {
+            let (_bytes, payload, files) =
```
germag: Could a short read be a problem, since `_bytes` is discarded?

vhost/src/vhost_user/mod.rs
```diff
@@ -52,8 +52,11 @@ mod backend_req;
 #[cfg(feature = "vhost-user-backend")]
 pub use self::backend_req::Backend;
 mod gpu_backend_req;
+mod gpu_frontend_req;
 pub mod gpu_message;
 pub use self::gpu_backend_req::GpuBackend;
+pub use self::gpu_frontend_req::GpuFrontend;
+pub use self::gpu_message::{GpuBackendReq, VirtioGpuRespDisplayInfo, VirtioGpuRespGetEdid};
```
germag: Why re-exporting these?

vhost/src/vhost_user/gpu_frontend_req.rs
dorindabassey: Right, FDs arrive with recv_header(). Simplified to use header files directly.

vhost/src/vhost_user/frontend.rs
dorindabassey: No, the existing backend handler for `GPU_SET_SOCKET` in `backend_req_handler.rs` doesn't check any protocol feature either. This matches QEMU's implementation.

vhost/src/vhost_user/gpu_frontend_req.rs
dorindabassey: Right, but this is a pre-existing pattern in `recv_into_buf()` itself. I can add the check in a seperate commit.

vhost/src/vhost_user/mod.rs
dorindabassey: an oversight, Fixed, thanks!

vhost/src/vhost_user/gpu_frontend_req.rs
dorindabassey: Done.

dorindabassey:
Hi @germag I have addressed the comments and rebased the PR. ready for review whenever you have the chance Thanks!

PR #374, opened 2026-08-20
Call backend event handler on vring enabled

jemoreira:
### Summary of the PR

There might already be available buffers in the vring when it is enabled. This is especially likely when enabling the vring after a live migration/restore from snapshot. Call the backend's handler immediately after enabling the vring to give it a chance to process those buffers.

The backend's handler is called only once if it is kicked and enabled at the same time, however it is still possible to call it when no available descriptors are present. This type of spurious calls are already possible and should be graciously handled by the backend implementation.

### Requirements

requirements:

  test.


jemoreira:
The only way I can think of to test this change is to run the `run` method in a thread and write to mock backend's exit event fd, but that depends on the fact that the exit event fd is processed after the events I added which is very brittle. Otherwise I'd have to refactor that run method so I can run a single iteration of the `'epoll` loop, which seemed like a too significant change to make.

rbradford:
In the VMM (CH) we trigger all the interrupts on restore. That seems the best way to handle this?

jemoreira:
> In the VMM (CH) we trigger all the interrupts on restore. That seems the best way to handle this?

This was a somewhat long discussion, the long version can be found in https://github.com/rust-vmm/vhost-device/pull/936 and https://lore.kernel.org/qemu-devel/20260604201029.250450-2-stefanha@redhat.com/. The gist is: the spec used to imply the VMM needed to interrupt the backends, but most backends were implemented in a different way than the spec mandated, so the spec was updated to match the implementations, which no longer requires the VMMs to kick on restore, while at the same time recommending VMMs do it anyways to support any backend implementation that followed the spec to the letter before the update.

So, while most VMMs do send those interrupts, it's not mandated by the spec and therefore backends can't rely on it.

jemoreira:
friendly ping

vhost-user-backend/Cargo.toml
```diff
@@ -19,6 +19,7 @@ postcopy = ["vhost/postcopy", "userfaultfd"]
 [dependencies]
 libc = "0.2.39"
 log = "0.4.17"
+nix = { version = "0.31", features = ["fs", "poll"] }
```
rbradford: I really don't want to pull in nix here. Can you find a solution using the libc / vmm-sys-util / epoll crates?

rbradford (changes requested):

vhost-user-backend/src/vring.rs
```diff
@@ -245,6 +250,21 @@ impl<M: GuestAddressSpace> VringState<M> {
...
+            let borrowed_kick = unsafe { BorrowedFd::borrow_raw(kick.as_raw_fd()) };
+            let mut pollfds = [PollFd::new(borrowed_kick, PollFlags::POLLIN)];
+            let nready = poll(&mut pollfds, PollTimeout::ZERO).map_err(Into::<io::Error>::into)?;
+            return Ok(nready > 0);
+        }
+
+        // A vring without a kick fd is always "kicked" since it must be polled for as long as it is
+        // active.
+        Ok(true)
+    }
```
rbradford: I think we can just tolerate sending a spurious extra notification which would simplify all this.

jemoreira: dropped it.

vhost-user-backend/Cargo.toml
jemoreira: removed it: was only needed to avoid extra notifications

vhost-user-backend/src/event_loop.rs
```diff
@@ -213,12 +266,37 @@ where
...
             .map_err(VringEpollError::HandleEventBackendHandling)?;
 
         Ok(false)
     }
+
+    fn handle_num_queues_event(&self, evset: EventSet) -> VringEpollResult<bool> {
+        // The num_queues event can be triggered by the exit event fd or a vring becoming enabled
+        match self.vring_enabled_event_fd.0.consume() {
+            Ok(()) => {
+                for device_event in &*self.vrings_with_pending_events.lock().unwrap() {
```
stefano-garzarella: From LLM reviewing, not sure if it's valid:
```
  - C1 (vhost-user-backend/src/event_loop.rs:285-293): handle_num_queues_event() iterates
    &*self.vrings_with_pending_events.lock().unwrap() but never removes the entries it just serviced. Since
    entries are only cleared by a real kick (handle_event(), line 270) or by unregister_event(), an
    enabled-but-not-yet-kicked vring stays in the set for the device's lifetime, and every subsequent
    vring-enable re-notifies all of them. I verified this empirically with a scratch test: enabling vring 0 →
    1 backend call; enabling vring 1 → 3 total; a third notify → 5 total. With the default queues_per_thread()
    (all queues on one thread) this is O(N²) spurious handle_event() calls per device bring-up. The set
    should be drain()ed (or std::mem::taken) instead of iterated by reference.
```

vhost-user-backend/src/event_loop.rs
```diff
@@ -102,12 +128,26 @@ where
             None
         };
 
+        let vring_enabled_event_fd =
+            new_event_consumer_and_notifier(EventFlag::NONBLOCK | EventFlag::CLOEXEC)
+                .map_err(VringEpollError::NewEventConsumerNotifier)?;
+        let vring_enabled_event = backend.num_queues() as u64;
```
stefano-garzarella: Why using the same `data` used for exit event?

I was hoping this would be explained more clearly in the commit message, which only describes what the patch does, not how or why...

vhost-user-backend/src/event_loop.rs
```diff
@@ -213,12 +266,37 @@ where
...
+                    if vring_idx >= self.backend.num_queues() {
+                        return Err(VringEpollError::VringIndexOutOfRange(vring_idx));
+                    }
+                    self.backend
+                        .handle_event(*device_event, evset, &self.vrings, self.thread_id)
+                        .map_err(VringEpollError::HandleEventBackendHandling)?;
+                }
+                Ok(false)
+            }
+            Err(e) if e.kind() == ErrorKind::WouldBlock => Ok(true),
```
stefano-garzarella, line 300: IMO this must be documented somewhere. IIUC this is working just because exit fd is not actaully drained, right?
It looks really fragile IMO.

stefano-garzarella: At least this should be explained in the commit description...

vhost-user-backend/src/event_loop.rs
```diff
@@ -213,12 +266,37 @@ where
             }
         }
 
+        // Reset the pending events mark on this vring since the backend is being notified.
+        self.vrings_with_pending_events
+            .lock()
+            .unwrap()
+            .remove(&device_event);
```
stefano-garzarella: Would it be too expensive to acquire this mutex for each event? Could we also use an AtomicBool to check if there is at least one pending event that needs to be reset, and if so, acquire the mutex?

Or maybe directly a Vec<AtomicBool> with a bool for each vring.

stefano-garzarella:
@jemoreira we usually ask PR authors to close comments that are addressed. If the commenter is unhappy, they will re-open it.

vhost-user-backend/src/event_loop.rs
jemoreira: > Would it be too expensive to acquire this mutex for each event?

(In Linux) Only when contended, otherwise lock and unlock are [`compare_exchange` on an atomic value](https://doc.rust-lang.org/src/std/sys/sync/mutex/futex.rs.html#25). The only possible sources of contention for this are `register_event` and `unregister_event`, which are called rarely so most of the time this will be uncontended and have the same cost as an atomic variable. `handle_num_queues_event` is called from the same thread as this function.

This is only here to avoid a possible double notification to the backend if a vring is enabled and kicked at roughly the same time. I can remove it if that's not a problem.

vhost-user-backend/src/event_loop.rs
jemoreira: It is valid, used `std::mem::take` to clear the set.

vhost-user-backend/src/event_loop.rs
jemoreira: Exit fd is never drained now because it was never drained without my change either. I don't know why that is (maybe because it's not created by this code and therefore not guaranteed to be an event fd?) so I didn't want to change it.

This is slightly different than the previous implementation because when that one got the event it would check `exit_event_fd.is_some()` to decide whether to exit or call `handle_event`. This implementation assumes the only reasons this event can be triggered for are the exit fd or vring enabled fd and never calls handle_event with num_queues as the device_idx because of [this comment](https://github.com/rust-vmm/vhost/blob/main/vhost-user-backend/src/event_loop.rs#L133).

Added more comments here and an explanation in the commit description.

rbradford (commented):
Since the backend implementation needs to be able to handle spurious events I think its completely reasonable to simplify this to just emit them whenever it might have been enabled. It's not like this will lead to an excessive amount of excess work and it would simplify the code significantly.

vhost-user-backend/src/event_loop.rs
stefano-garzarella: (sorry, this can be left unresolved, since we are still discussing)

> This is only here to avoid a possible double notification to the backend if a vring is enabled and kicked at roughly the same time. I can remove it if that's not a problem.

IMO we can live with this spurious notifications. I guess the chance to happen is very low, but even if that happen, nothing can go wrong, right?

jemoreira:
> Since the backend implementation needs to be able to handle spurious events I think its completely reasonable to simplify this to just emit them whenever it might have been enabled. It's not like this will lead to an excessive amount of excess work and it would simplify the code significantly.

done

vhost-user-backend/src/event_loop.rs
```diff
@@ -219,6 +272,35 @@ where
...
+                        .map_err(VringEpollError::HandleEventBackendHandling)?;
+                }
+                // There may have been a notification in the exit fd too, but it wasn't consumed and
+                // it's not safe to attempt to because that fd may not be non-blocking. Returning
+                // false ensures there will be another iteration of the main loop which will trigger
+                // the same event if exit fd is readable.
+                Ok(false)
+            }
+            // With no notification in the vring enabled fd the only other option is the exit fd.
+            Err(e) if e.kind() == ErrorKind::WouldBlock => Ok(true),
```
stefano-garzarella: To keep the previous behaviour, should this be:

```suggestion
            Err(e) if e.kind() == ErrorKind::WouldBlock => Ok(self.exit_event_fd.is_some()),
```

vhost-user-backend/src/event_loop.rs
```diff
@@ -219,6 +272,35 @@ where
...
+        match self.vring_enabled_event_fd.0.consume() {
+            Ok(()) => {
+                let vrings_with_pending_events =
+                    std::mem::take(&mut *self.vrings_with_pending_events.lock().unwrap());
+                for device_event in vrings_with_pending_events {
+                    let vring_idx = device_event as usize;
+                    if vring_idx >= self.backend.num_queues() {
+                        return Err(VringEpollError::VringIndexOutOfRange(vring_idx));
+                    }
+                    self.backend
```
stefano-garzarella, line 290: Should we also check that the vring is still enabled, like we do in `handle_event` after calling `vring.read_kick()`?

vhost-user-backend/src/event_loop.rs
```diff
@@ -139,11 +179,24 @@ where
     }
 
     pub(crate) fn register_event(&self, fd: RawFd, ev_type: EventSet, data: u64) -> Result<()> {
+        if data < self.backend.num_queues() as u64 {
+            self.vrings_with_pending_events
+                .lock()
+                .unwrap()
+                .insert(data as u16);
+            self.vring_enabled_event_fd.1.notify()?;
+        }
```
stefano-garzarella, line 190: From LLM:
```
  - M3 (vhost-user-backend/src/event_loop.rs:182-188): The notify is sent before
    epoll.ctl(Add) and is never rolled back. Two consequences: (a) the backend can be
    told to process a vring whose kick fd is not yet in the epoll set; (b) when
    epoll.ctl fails, the index stays in the pending set and the notification stands.
    Case (b) includes the AlreadyExists path that handler.rs:229-233 explicitly treats
    as normal ("asked by the frontend to enable an already enabled queue"), so a
    frontend re-sending SET_VRING_ENABLE 1 repeatedly generates a backend handle_event
    each time.
```

Should we just put this after the epoll.clt() call?

vhost-user-backend/src/event_loop.rs
```diff
@@ -219,6 +272,35 @@ where
...
+        // The num_queues event can be triggered by the exit event fd or a vring becoming enabled.
+        // The vring_enabled_event_fd is guaranteed to be non-blocking so reading from it will
+        // determine which fd triggered the event.
+        match self.vring_enabled_event_fd.0.consume() {
+            Ok(()) => {
+                let vrings_with_pending_events =
+                    std::mem::take(&mut *self.vrings_with_pending_events.lock().unwrap());
+                for device_event in vrings_with_pending_events {
+                    let vring_idx = device_event as usize;
+                    if vring_idx >= self.backend.num_queues() {
```
stefano-garzarella: From LLM:
```
  - S2 (vhost-user-backend/src/event_loop.rs:286): The bounds check uses
    self.backend.num_queues(), but the value indexes self.vrings, which for
    multi-threaded backends is a per-thread subset (handler.rs:131-136). The check is
    therefore weaker than the one in handle_event (line 257, < self.vrings.len()) and
    does not actually protect the backend's indexing. 
```

vhost-user-backend/src/event_loop.rs
jemoreira: done

vhost-user-backend/src/event_loop.rs
jemoreira: checked index against `self.vrings.len()` instead.

vhost-user-backend/src/event_loop.rs
jemoreira: It wouldn't hurt, but I don't think it would help either: If the vring can become disabled between the notification being sent and this calling `backend.handle_event()` it may also become disabled between the check for enabled and the call to `backend.handle_event()`.

Also: this references `VringT` which, unlike `VringState`, doesn't have an `is_enabled` method. It would be easy to add it, but I don't know if it's ok to expose it publicly in that trait. Please let me know if I should add the function to the trait and check it here.

vhost-user-backend/src/event_loop.rs
jemoreira: done

vhost-user-backend/src/event_loop.rs
stefano-garzarella: ack

stefano-garzarella (approved):

stefano-garzarella:
@rbradford can you take a look at this?

rbradford (approved):

PR #377, opened 2026-09-21
vhost-user-backend: add a hook for stopped vrings

daprilik:
### Summary of the PR

Turns out marking a vring not ready does not necessarily mean that every request previously accepted by an asynchronous backend has completed. `GET_VRING_BASE` could therefore return the available-ring position while the backend was still publishing used entries for that queue.

Add a default `stop_vring()` hook after the queue is marked not ready and unregistered, but before its position is read. This gives asynchronous backends a place to drain outstanding work, while preserving the existing behavior for synchronous implementations.

The `Mutex` and `RwLock` adapters invoke the hook while holding their backend lock. Implementations must ensure that their drain can make progress without reacquiring it.

### Requirements

requirements:

  test.


sboeuf (dismissed):
LGTM

rbradford:
@daprilik Need to rebase but lgtm

daprilik:
Rebased, should be gtg 🤞

rbradford (approved):

sboeuf (approved):

PR #378, opened 2026-09-22
vhost-user: allow an fd with REM_MEM_REG

rbradford:
Some frontends attach an fd to REM_MEM_REG. Accept it instead of
rejecting the message as invalid.

Signed-off-by: Rob Bradford <rbradford@meta.com>

### Summary of the PR

### Requirements

requirements:

  test.


sboeuf (dismissed):

rbradford:
@stefano-garzarella Can you PTAL.

vhost/src/vhost_user/backend_req_handler.rs
```diff
@@ -951,6 +951,7 @@ impl<S: VhostUserBackendReqHandler> BackendReqHandler<S> {
                 | FrontendReq::SET_BACKEND_REQ_FD
                 | FrontendReq::SET_INFLIGHT_FD
                 | FrontendReq::ADD_MEM_REG
+                | FrontendReq::REM_MEM_REG
```
stefano-garzarella, line 959: Can we add a comment here why we are doing this?

Maybe we can cite the spec:

> No file descriptors SHOULD be passed in the ancillary data. For compatibility with existing incorrect implementations, the back-end MAY accept messages with one file descriptor. If a file descriptor is passed, the back-end MUST close it without using it otherwise.


I guess we are already doing the second part, I mean we are closing it, right?

rbradford: I'm confused. Where would you want to cite the spec? Isn't that a bit overkill for just this? This was only missing because this crate itself doesn't generate this but QEMU does.

stefano-garzarella: While reviewing, my question was, why `check_attached_files()` should return Ok() if `REM_MEM_REG` is not expecting any fd in the messsage?

I didn't find the answer neither in the code change, neither in the commit message, but I had to go into the spec and check the reason (IIUC some fronted, maybe QEMU, is passing an fd, but it should not do that). I'd like to avoid these steps in the future...

rbradford: I got it. Do you think a comment is appropriate or maybe it would be better in the commit instead?

stefano-garzarella: Not a strong opinion, I would put it in the code so it's definitely clear if anyone were to ask the same question, but if you'd rather just leave it in the commit message, that's fine too.

rbradford: I understand - I did both!

stefano-garzarella (approved):

sboeuf (changes requested):
LGTM except the `back-end` terminology which should be `backend`.

rbradford:
> LGTM except the `back-end` terminology which should be `backend`.

That's a direct quote from the spec. Take if up with the QEMU folks! :-)

sboeuf:
> LGTM except the `back-end` terminology which should be `backend`.

Ah nevermind, the spec uses the same terminology.

sboeuf (approved):

stefano-garzarella:
> > LGTM except the `back-end` terminology which should be `backend`.
> 
> That's a direct quote from the spec. Take if up with the QEMU folks! :-)

yeah, IMO is better to cite the spec as is...

rbradford:
@stefano-garzarella Can you retry the buildkite?

rbradford:
Looks "up-to-date" required for merging? That's annoying as it requires re-reviews as it gets rid of them when you update.

PR #379, opened 2026-09-22
vhost-user-backend: avoid backend lock during registration

rogervn:
### Summary of the PR

VringEpollHandler::register_event() and unregister_event() can be
called while update_vring_registration() holds a vring lock. Calling
backend.num_queues() from there acquires the backend adapter lock.

Event handling can acquire these locks in the opposite order: backend
first, then vring. This can cause a deadlock.

This change stores the number of queues from the backend prior to
calling register_event and unregister_event and pass that as a
parameter to prevent that.

We also add a regression test that reproduced this on the former
code and validates the fix.

Signed-off-by: Rogerio Vinhal Nunes <rogerio@vnunes.win>

### Requirements

requirements:

- [X] All commits in this PR have Signed-Off-By trailers (with
- [X] All added/changed functionality has a corresponding unit/integration
  test.
- [X] All added/changed public-facing functionality has entries in the "Upcoming 
- [X] Any newly added `unsafe` code is properly documented.


vhost-user-backend/src/event_loop.rs
```diff
@@ -84,6 +84,7 @@ pub type VringEpollResult<T> = std::result::Result<T, VringEpollError>;
 pub struct VringEpollHandler<T: VhostUserBackend> {
     epoll: Epoll,
     backend: T,
+    num_queues: usize,
```
sboeuf, line 87: Isn't this redundant information being stored since we already have `vrings` and we should be able to check `vrings.len()` to get that same info, right?

Or is it because we expect the backend to possibly activate less queues than the amount of available queues? So maybe call it `num_active_queues` or `backend_queues`?

sboeuf (changes requested):
Could you please write a clear commit message to explain what you're trying to solve here?

vhost-user-backend/src/event_loop.rs
rbradford: Driver can definitely activate fewer queues than advertised. I don't think a device can choose to activate fewer queues but if the driver activates fewer queues is that visible in that value?

rbradford (commented):
@rogervn Please take a look at the style of commit messages in the repo and try to follow that e.g. should have prefix "vhost-user-backend: ..."

vhost-user-backend/src/event_loop.rs
```diff
@@ -84,6 +84,7 @@ pub type VringEpollResult<T> = std::result::Result<T, VringEpollError>;
...
     vrings: Vec<T::Vring>,
```
rbradford, line 88: So `vrings` can be a subset of queues as the queues can be partitioned across multiple worker (ie. threads) so we do need the actual number of queues.

sboeuf: Okay, well then let's make sure we use a very comprehensive name.

rbradford: `num_backend_queues` seems fine to me. But i was happy with the original `num_queues`.

sboeuf: I think I prefer `num_backend_queues`.

stefano-garzarella:
CC @jemoreira

vhost-user-backend/src/event_loop.rs
rogervn: No worries, changed it `num_backend_queues`.

vhost-user-backend/src/event_loop.rs
rogervn: Yeah, IIUC `vrings` has the queues assigned to this worker whilst `backend` has the total queue count for the event.

rbradford (approved):

sboeuf (approved):

stefano-garzarella:
Next time, can we give time to people CCed to review it? Especially when it's a fix of an issue recently introduced.

